PR #42, opened 2014-09-26
Bump upper bounds in vector-benchmarks.cabal

mgsloan:
Also:

+ remove fixed version constraint for vector

+ normalized whitespace


mgsloan:
Clearly this will never get merged, so closing.

Is vector no longer maintained?  There are many outstanding PRs

PR #44, opened 2014-09-27
Speed up index check slightly

treeowl:
Previously, the index was checked using `0 <= i && i < n`. Since `&&` short-circuits, this compiles to signed-compare/branch/signed-compare. It's slightly better to use unsigned comparison, which takes it down to a single unsigned-compare operation. This technique relies on `n >= 0`.


cartazio:
the design idea here is *good*, but changing the internal rep to use Word for the comparison scares me.

also do we have any bechnmarks that branch prediction fails to just work? (and or that the Int# style rep on newer ghc's wouldnt suffice?)

PR #53, opened 2014-11-12
Replace macros with ConstraintKinds (fixes #48)

### LINKED ISSUE #48: replace macrollogy in test suite with constraint kinds
https://github.com/haskell/vector/blob/master/tests/Tests/Vector.hs#L27-L34

```
#define COMMON_CONTEXT(a, v) \
VANILLA_CONTEXT(a, v), VECTOR_CONTEXT(a, v)
#define VANILLA_CONTEXT(a, v) \
Eq a, Show a, Arbitrary a, CoArbitrary a, TestData a, Model a ~ a, EqTest a ~ Property
#define VECTOR_CONTEXT(a, v) \
Eq (v a), Show (v a), Arbitrary (v a), CoArbitrary (v a), TestData (v a), Model (v a) ~ [a], EqTest (v a) ~ Property, V.Vector v a
```

vile stuff, lets kill it!

Shimuuar:


cartazio:
@Shimuuar  any interest in resuscitating this or some other effort to cleanup the test suite code?I'm happy to help shepard any such work :)

Shimuuar:
What are supported GHC versions? Earlier this day fix for GHC 7.0 was pushed and it doesn't have constraint kinds.

cartazio:
@Shimuuar we're dropping 7.0 soon (@dolio?)
I presume that means we should consider moving to only supporting >= 7.4 or 7.6 for the next release? thoughts @dolio ?

cartazio:
either way, if you'd be interested in generally trying to make the test suite codes less terrible please join in :)

dolio:
I was thinking about just dropping 7.0 (especially since I thought it was panicking, but that was only 7.0.1), but then I realized I didn't have to for that one fix. But really it'd be nice to just use newer features, especially for stuff like this.

I guess we probably don't have anyone watching this who will argue that 7.2 and earlier are important?

cartazio:
7.2 was explicitly a preview release that had "test for bugs but don't use
for serious business" on the tin.

The oldest industrial code I'm aware of that's being actively maintained
uses 7.8.

A supporting 4 major versions policy would still put the min bound at 7.6.

I can't imagine anyone who'd object to us moving min version bound up to
the version corresponding to ghc 7.4 in master. Or even 7.8/7.6 if we want
to be edgy.

If we have due cause for fancier stuff I'm all for it, but I'm pretty
comfortable presuming 7.4 as the new min version is kosher with humanity
and Haskelldom

PR #74, opened 2015-01-29
add mapMaybe and imapMaybe

chadaustin:
This is my first pull Haskell pull request, so please review closely.  :)
I would find these functions highly convenient, especially in scenarios like `filterMap Json.decode`

Also, did I do the tests correctly?  Should the matrix be filled out with monadic equivalents too?  Is the documentation duplication OK?

### DIFF
--- a/Data/Vector.hs
+++ b/Data/Vector.hs
@@ -112,7 +112,9 @@ module Data.Vector (
   -- * Working with predicates
 
   -- ** Filtering
-  filter, ifilter, filterM,
+  filter, ifilter,
+  filterMap, ifilterMap,
+  filterM,
   takeWhile, dropWhile,
 
   -- ** Partitioning
@@ -1078,6 +1080,16 @@ filter = G.filter
 ifilter :: (Int -> a -> Bool) -> Vector a -> Vector a
 {-# INLINE ifilter #-}
 ifilter = G.ifilter
+
+-- | /O(n)/ Drop elements when predicate returns Nothing
+filterMap :: (a -> Maybe b) -> Vector a -> Vector b
+{-# INLINE filterMap #-}
+filterMap = G.filterMap
+
+-- | /O(n)/ Drop elements when predicate, applied to index and value, returns Nothing
+ifilterMap :: (Int -> a -> Maybe b) -> Vector a -> Vector b
+{-# INLINE ifilterMap #-}
+ifilterMap = G.ifilterMap
 
 -- | /O(n)/ Drop elements that do not satisfy the monadic predicate
 filterM :: Monad m => (a -> m Bool) -> Vector a -> m (Vector a)
--- a/Data/Vector.hs
+++ b/Data/Vector.hs
@@ -113,7 +113,7 @@ module Data.Vector (
 
   -- ** Filtering
   filter, ifilter,
-  filterMap, ifilterMap,
+  mapMaybe, imapMaybe,
   filterM,
   takeWhile, dropWhile,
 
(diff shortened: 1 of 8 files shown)


cartazio:
a quick skim it looks pretty solid, have you sanity checked that it fuses properly on a few toy examples? :)

cartazio:
maybe you could add a mini benchmark or so to the benchmark suite?

Shimuuar:
These functions could be generalized to Foldable f => (a -> f b) -> v a ->
v b. I however not sure that such generic function could be optimized nicely
30 Янв 2015 г. 4:01 пользователь "Carter Tazio Schonwald" <
notifications@github.com> написал:

takano-akio:
Would `mapMaybe` be a better name for `filterMap`? This is how the operation is called in `Data.Maybe`, `Data.Map.Lazy` and other modules.

chadaustin:
mapMaybe is a fine name.  Consistency is great, thanks for the heads up.  I also think @cartazio is right that this needs a benchmark to prove that its implementation shouldn't just be filter/fold before it's merged.

cartazio:
dont you mean `mapMaybe f  = map (fromJust .f) . filter (isJust . f)`? Or is there a more clever approach using fold i'm not understanding?

chadaustin:
I tried writing a benchmark, but after an hour of struggling with cabal, I gave up.  You know, the typical errors.  Either way, I'll rename it to filterMap...

Perhaps the benchmarks should be moved into the root .cabal file.

```
chad-mbpro ~/projects/vector/benchmarks $ cabal install --only-dependencies
Resolving dependencies...
cabal: Could not resolve dependencies:
trying: vector-0.11.0.0/installed-d33... (user goal)
trying: vector-benchmarks-0.10.9 (user goal)
next goal: criterion (dependency of vector-benchmarks-0.10.9)
rejecting: criterion-1.0.2.0, 1.0.1.0, 1.0.0.2, 1.0.0.1, 1.0.0.0, 0.8.1.0,
0.8.0.2, 0.8.0.1, 0.8.0.0 (conflict: vector-benchmarks => criterion>=0.5 &&
<0.7)
trying: criterion-0.6.2.1
next goal: vector-algorithms (dependency of criterion-0.6.2.1)
rejecting: vector-algorithms-0.6.0.3, 0.6.0.2, 0.6.0.1, 0.5.4.2 (conflict:
vector==0.11.0.0/installed-d33..., vector-algorithms => vector>=0.6 && <0.11)
rejecting: vector-algorithms-0.5.4.1, 0.5.4, 0.5.3.1, 0.5.3 (conflict:
vector==0.11.0.0/installed-d33..., vector-algorithms => vector>=0.6 && <0.10)
rejecting: vector-algorithms-0.5.2, 0.5.1, 0.5.0, 0.4 (conflict:
vector==0.11.0.0/installed-d33..., vector-algorithms => vector>=0.6 && <0.8)
rejecting: vector-algorithms-0.3.4, 0.3.3, 0.3.2, 0.3.1, 0.3 (conflict:
criterion => vector-algorithms>=0.4)
Dependency tree exhaustively searched.

Note: when using a sandbox, all packages are required to have consistent
dependencies. Try reinstalling/unregistering the offending packages or
recreating the sandbox.
```

chadaustin:
Renamed filterMap to mapMaybe.  Thanks for the suggestion.  It matches the mapMaybe functions in Data.Maybe, Data.IntMap, and Data.Map.  Even if map fromJust . filter isJust fuses as well, I still think this function is worth having for convenience.

But ultimately that's up to you all.  :)

PR #78, opened 2015-04-24
added monadzip instance for Data.Vector

fhaust:
Adding `MonadZip` instances for `Vector` types allows to use `-XMonadComprehensions` and `-XParallelComprehensions` together:

    >>> :set -XMonadComprehensions
    >>> :set -XParallelListComp
    >>> import qualified Data.Vector as V
    >>> let as = V.fromList [0..10]
    >>> let bs = V.fromList [0..10]
    >>> [ a + b | a <- as | b <- bs ]
    fromList [0,2,4,6,8,10,12,14,16,18,20]

I would like to add the instance for all other types (especially the unboxed type) but I see that there are no instances of `Monad` nor `MonadPlus` there which is probably for a reason?

### DIFF
--- a/Data/Vector.hs
+++ b/Data/Vector.hs
@@ -308,6 +308,8 @@ instance MonadPlus Vector where
   {-# INLINE mplus #-}
   mplus = (++)
 
+-- MonadZip was added in base-4.4.0
+#if MIN_VERSION_base(4,4,0)
 instance MonadZip Vector where
   {-# INLINE mzip #-}
   mzip = zip
@@ -317,6 +319,7 @@ instance MonadZip Vector where
 
   {-# INLINE munzip #-}
   munzip = unzip
+#endif
 
 instance Applicative.Applicative Vector where
   {-# INLINE pure #-}
--- a/Data/Vector.hs
+++ b/Data/Vector.hs
@@ -166,6 +166,7 @@ import Control.DeepSeq ( NFData, rnf )
 import Control.Monad ( MonadPlus(..), liftM, ap )
 import Control.Monad.ST ( ST )
 import Control.Monad.Primitive
+import Control.Monad.Zip
 
 import Prelude hiding ( length, null,
                         replicate, (++), concat,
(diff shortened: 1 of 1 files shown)


gregorycollins:
That module was added in base 4.4; please add ifdefs making compilation conditional on having at least that version.

cartazio:
@gregorycollins  should we still support ghc 7.0 with vector HEAD? 

@fhaust this looks like a reasonable patch! have you tried to verified that whatever laws of MonadZip hold? (it looks like they do, but probably worth checking!)

You wont be able to write such an instance for Storable or Unboxed vectors, because unliked Boxed vector, they have class constraints on their element type! (Storable and Unbox) respectively.  So dont fret about not having those instances, they're not expressible with the standard monad type classes we have in the base libraries

gregorycollins:
@cartazio: there's little reason not to support it if it's not too bothersome to do so, and 7.0 is still being built on the Travis CI.

cartazio:
fair enough :) 
(i do think that as ghc matures more, we should explore new ways to improve the vector library, which could hypothetically push for a smaller support window, though not necessarily)

fhaust:
@cartazio added CPP directives

About the laws, I tested them "by hand". But that's probably better checked by quickcheck. Not sure how to add those though.

cartazio:
you may wanna look in the test suite :) 
https://github.com/haskell/vector/blob/master/tests/Tests/Vector.hs#L539-L551 might be the right spot to add some boxed vector specific tests perhaps?

fhaust:
@cartazio yep ... I looked at [lines 491-496](https://github.com/haskell/vector/blob/master/tests/Tests/Vector.hs#L491-L496 before) before, but those don't look like the quickcheck tests I am used to. I was hesitant to add something that was "out of line" with the tests there.

cartazio:
@fhaust hey, we're on track to do some test suite cleanup in master finally, any comfort in at least updating this PR to provide some quickcheck properties we can wire in? :)

fhaust:
​I'd love to, but last time I tried, I found `vectors` QuickCheck code to
be pretty non-standard. I just didn't manage to implement the properties
there.​


Am 24.07.2016 9:41 nachm. schrieb "Carter Tazio Schonwald" <
notifications@github.com>:

cartazio:
Well, the good news is we plan on burning the test suite with fire. So just write some vanilla properties

fhaust:
Is this already in effect? If so where do the new property tests go?

cartazio:
Nope. We need help doing the test suite sanity ;)

fhaust:
Created a new [pull request](https://github.com/haskell/vector/pull/129) ... that one includes test cases.

cartazio:
yay!

PR #81, opened 2015-05-16
add  Pure :: (* -> * -> *) -> (* -> *) type family alongside the stan…

cartazio:
…dard Mutable type family.

This is to allow better bidirectional inference in code that mixes Mutable and Pure Vectors.

also the first of a few different changes / additions we agreed on doing for vector 0.11

theres a few few style questions that should probably be discussed before this gets merged in, such as my choice in using a CPP approach as shorthand in modules that already were doing CPP hacks.

@dolio  @tibbe  @erikd  et al would love some feedback please :)

### DIFF
--- a/Data/Vector/Fusion/Bundle/Monadic.hs
+++ b/Data/Vector/Fusion/Bundle/Monadic.hs
@@ -107,7 +107,7 @@ import Data.Word ( Word8, Word16, Word32, Word, Word64 )
 #include "vector.h"
 #include "MachDeps.h"
 
-data Chunk v a = Chunk Int (forall m. (PrimMonad m, Vector v a) => Mutable v (PrimState m) a -> m ())
+data Chunk v a = Chunk Int (forall m mv. (PrimMonad m, Vector v a, MUTABLE_PURE(mv,v)) => mv (PrimState m) a -> m ())
 
 -- | Monadic streams
 data Bundle m v a = Bundle { sElems  :: Stream m a
--- a/Data/Vector/Generic/Base.hs
+++ b/Data/Vector/Generic/Base.hs
@@ -35,7 +35,8 @@ type family Pure (mv :: * -> * -> *) :: * -> *
 -- | Class of immutable vectors. Every immutable vector is associated with its
 -- mutable version through the 'Mutable' type family. Methods of this class
 -- should not be used directly. Instead, "Data.Vector.Generic" and other
--- Data.Vector modules provide safe and fusible wrappers.
+-- Data.Vector modules provide safe and fusible wrappers. Every 'Vector' instance
+-- must provide instances of both 'Mutable' and 'Pure' that jointly obey @v ~ 'Pure' ('Mutable' v)@.
 --
 -- Minimum complete implementation:
 --
--- a/Data/Vector/Generic/Base.hs
+++ b/Data/Vector/Generic/Base.hs
@@ -15,7 +15,7 @@
 --
 
 module Data.Vector.Generic.Base (
-  Vector(..), Mutable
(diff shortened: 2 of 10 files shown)


cartazio:
@ekmett i know this is a PR you've wanted for a long time :)

cartazio:
hrm, my ` (v ~ Pure (Mutable v), MVector (Mutable v) a)` contraint on the Vector class would mean we'd have to drop ghc 7.0  and 7.2 support, but that constraint is needed to give (newer) ghc enough info to actually type check stuff.

cartazio:
if this PR gets accepted, we'll need to raise the lowerbounds on base / ghc

cartazio:
the ghc head build is giving some interesting warning / feedback 
eg 

```Data/Vector/Fusion/Bundle.hs:114:1: warning:
    Tab character found here.
    Please use spaces instead.
```

cartazio:
that said, i'm not sure if the notion of redundant constraints is correct, eg 
```
Data/Vector/Generic.hs:708:11: warning:
    Redundant constraint: v ~ Pure mv
    In the type signature for:
       create :: (Vector v a, v ~ Pure mv, mv ~ Mutable v) =>
                 (forall s. ST s (mv s a)) -> v a
```

if the `v ~ Pure mv` constraint were dropped, then given only `mv`, we'd not be able to deduce `v`! Or am I overlooking something

Data/Vector/Fusion/Bundle/Monadic.hs
```diff
@@ -107,7 +107,7 @@ import Data.Word ( Word8, Word16, Word32, Word, Word64 )
 #include "vector.h"
 #include "MachDeps.h"
 
-data Chunk v a = Chunk Int (forall m. (PrimMonad m, Vector v a) => Mutable v (PrimState m) a -> m ())
+data Chunk v a = Chunk Int (forall m mv. (PrimMonad m, Vector v a, MUTABLE_PURE(mv,v)) => mv (PrimState m) a -> m ())
```
cartazio: i'm not sure if this change makes any difference what soever, but it seemed like a good move from a consistency perspective

cartazio:
it looks like adding `class (v ~ Mutable (Pure v)) => MVector v a where` to `MVector` will force all the pairs of MVector/Vector instances to be in shared internal modules

cartazio:
hrm, i dont understand the choice of pieces of code where ghc head trips over redundant constraint warnings. i'll dig into that more when i'm done travelling this weekend

cartazio:
@dolio we never did add this did we? lets revisit this for 0.12 and i'll clean it up

Data/Vector/Generic/Base.hs
```diff
@@ -45,19 +50,19 @@ type family Mutable (v :: * -> *) :: * -> * -> *
...
 --
-class MVector (Mutable v) a => Vector v a where
+class (v ~ Pure (Mutable v),MVector (Mutable v) a) => Vector v a where
   -- | /Assumed complexity: O(1)/
   --
   -- Unsafely convert a mutable vector to its immutable version
   -- without copying. The mutable vector may not be used after
   -- this operation.
-  basicUnsafeFreeze :: PrimMonad m => Mutable v (PrimState m) a -> m (v a)
+  basicUnsafeFreeze :: (PrimMonad m,mv~Mutable v,v~Pure mv) => mv (PrimState m) a -> m (v a)
```
OlivierSohn: Could `MUTABLE_PURE ` be used here?

lehins:
If I understand correctly this PR was trying to emulate injectivity that was implemented in #160 
Closing. Feel free to reopen if you think I am misunderstood this PR

PR #84, opened 2015-07-06
Avoid giving out uninitialized memory from safe vector creation functions.

dolio:


cartazio:
a first pass read, it look pretty solid!

roughly, the initcode is just doing a set 0 on the underlying byte array, whether its via Primitive, or via Foreign Pointer?

dolio:
Yes. Except for boxed vectors where it's a no-op.

For a moment last night, I thought this already existed as `clear`, but it turns out that is the exact opposite of what is needed (resets boxed vectors to undefined, no-op on unboxed vectors).

PR #89, opened 2015-07-30
Add 'createT', allowing monadic initialisation of any Traversable con…

nshepperd:
…tainer of vectors.

The current definition of `create`

    create :: Vector v a => (forall s. ST s (Mutable v s a)) -> v a 

is a little bit restrictive, as it does not allow us to return any other additional information calculated during the construction of the vector, such as vector indices. We can instead use a combinator like `create' :: Vector v a => (forall s. ST s (b, Mutable v s a)) -> (b, v a)` to return extra data alongside the vector, but I attempted to generalise this even further.

My formula right now is:

    createT :: (Traversable f, Vector v a) => (forall s. ST s (f (Mutable v s a))) -> f (v a)
    createT p = runST (p >>= traverse unsafeFreeze)

I believe this should be safe, as `runST (p >>= traverse freeze)` is safe and parametricity guarantees that the traversal cannot run any other ST actions (that would, say, write to the mutable vector) after the freeze.

The Traversable context allows for things such as returning auxiliary data (`instance Traversable ((,) b)`), returning multiple vectors (Traversable instances for lists, vectors, etc), even Trees of vectors, and a few other less useful seeming instances like Maybe.

I'm not sure of a few things: is it possible to integrate this with the stream fusion framework in any way (eg. with New)? Also, I added a very limited test, `prop_createT`, but I'm not clear on how meaningful that one actually is.

It would be great to have a pair of eyes run over this, or thought as to whether or not this would be useful and/or worth including in the library :)

### DIFF
--- a/Data/Vector.hs
+++ b/Data/Vector.hs
@@ -56,7 +56,7 @@ module Data.Vector (
   empty, singleton, replicate, generate, iterateN,
 
   -- ** Monadic initialisation
-  replicateM, generateM, create,
+  replicateM, generateM, create, createT,
 
   -- ** Unfolding
   unfoldr, unfoldrN,
@@ -686,6 +686,11 @@ create :: (forall s. ST s (MVector s a)) -> Vector a
 {-# INLINE create #-}
 -- NOTE: eta-expanded due to http://hackage.haskell.org/trac/ghc/ticket/4120
 create p = G.create p
+
+-- | Execute the monadic action and freeze the resulting vectors.
+createT :: Traversable f => (forall s. ST s (f (MVector s a))) -> f (Vector a)
+{-# INLINE createT #-}
+createT p = G.createT p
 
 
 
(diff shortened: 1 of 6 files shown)


dolio:
Sorry about not getting to this very soon.

My main concern was whether it is okay to call `unsafeFreeze` on an array more than once, because that could happen with the multi-element `Traversable` instances. However, it seems to be fine.

I think it's pretty unlikely that this will be possible to incorporate into New or the like in an efficient way. But it's probably worth having as a function.

nshepperd:
Ah, good catch, I didn't see that. So it's safe to call unsafeFreeze on the same vector multiple times?

cartazio:
idea: 

```
data New v a = New (forall s. ST s (Mutable v s a))

create :: (forall s. ST s (Mutable v s a)) -> New v a
{-# INLINE create #-}
create p = New p
```
and likewise 

```
type family Mutable (v :: * -> *) :: * -> * -> *
```


Question: would you be able to accomplish stuff using 
`mapM :: Monad m => (a -> m b) -> Bundle v a -> Bundle m v b`

this seems to be the traverse you might be looking for?

PR #90, opened 2015-07-30
Document mutable-copy's arguments

Peaker:
In `unsafeCopy` and others, the target/source were distinguished with haddock parameter docs.

But in `copy` the parameters are missing haddocks and are indistinguishable.


cartazio:
@dolio  thoughts?

PR #93, opened 2015-08-17
Fix GHC 7.10 warnings.

23Skidoo:


Data/Vector/Fusion/Bundle/Monadic.hs
```diff
@@ -888,6 +896,7 @@ enumFromTo_big_word x y = x `seq` y `seq` fromStream (Stream step x) (Exact (len
                         :: Monad m => Integer -> Integer -> Bundle m v Integer   #-}
 
 
+#if WORD_SIZE_IN_BITS > 32
 
 -- FIXME: the "too large" test is totally wrong
```
cartazio: why do we have a fixme in vector on "to large test?!" @dolio

PR #94, opened 2015-08-20
Bump upperbound for QuickCheck

phadej:


cartazio:
i've a branch for moving up to 2.9.* so the tests can drop orphans, thanks! please keep up the great suggestions and sorry about the latency :)

Issue #75, opened 2015-02-06
Alignment isn't respected in Data.Vector.Storable.


danielwaterworth:
I have a type where the alignment of the type is greater than 8 bytes. When I create a vector, using `fromList` for example, I find that the alignment isn't respected.

I'm using version 0.10.12.2.

dolio:
What does this mean exactly? Is it aligning to 8 and overwriting things?

DanielWaterworth:
It means the items aren't aligned in memory as they should be, which means peeking and pokeing may produce undefined behaviour.

cartazio:
Could you share your storable instance code?

Shimuuar:
I guess it could happen if `sizeof a` is not multiple of `alignment`. Vector uses `peek/pokeElemOff` which calculates offset in array as `n * sizeof a` so alignment isn't be respected in this case. Storable's documentation doesn't say much about alinment so I think this is main problem

DanielWaterworth:
Here you go:

    data Word512 =
      Word512
        {-# UNPACK #-} !Word64
        {-# UNPACK #-} !Word64
        {-# UNPACK #-} !Word64
        {-# UNPACK #-} !Word64
        {-# UNPACK #-} !Word64
        {-# UNPACK #-} !Word64
        {-# UNPACK #-} !Word64
        {-# UNPACK #-} !Word64
          deriving Show

    instance Storable Word512 where
      sizeOf _ = 64

      alignment _ = 64

      peek ptr =
        let
          ptr' = castPtr ptr
        in
          Word512 <$>
            peek ptr' <*>
            peek (plusPtr ptr' 8) <*>
            peek (plusPtr ptr' 16) <*>
            peek (plusPtr ptr' 24) <*>
            peek (plusPtr ptr' 32) <*>
            peek (plusPtr ptr' 40) <*>
            peek (plusPtr ptr' 48) <*>
            peek (plusPtr ptr' 56)

      poke ptr (Word512 a b c d e f g h) = do
        let ptr' = castPtr ptr
        poke ptr' a
        poke (plusPtr ptr' 8) b
        poke (plusPtr ptr' 16) c
        poke (plusPtr ptr' 24) d
        poke (plusPtr ptr' 32) e
        poke (plusPtr ptr' 40) f
        poke (plusPtr ptr' 48) g
        poke (plusPtr ptr' 56) h

cartazio:
[...]

PR #96, opened 2015-08-26
Make storable vectors respect alignment of underlying elements

sergv:
Fix for issue #75.

The only nontrivial bit is the use of `touch` in ForeignPtr finalizers for GHCs up to 7.6. It's needed because after transforming mutable byte array into pointer there're no references left from that pointer to original byte array because pointer is just an address.

### DIFF
--- a/Data/Vector/Fusion/Bundle.hs
+++ b/Data/Vector/Fusion/Bundle.hs
@@ -111,7 +111,7 @@ type Bundle = M.Bundle Id
 type MBundle = M.Bundle
 
 inplace :: (forall m. Monad m => S.Stream m a -> S.Stream m b)
-	-> (Size -> Size) -> Bundle v a -> Bundle v b
+        -> (Size -> Size) -> Bundle v a -> Bundle v b
 {-# INLINE_FUSED inplace #-}
 inplace f g b = b `seq` M.fromStream (f (M.elements b)) (g (M.size b))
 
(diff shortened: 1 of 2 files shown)


cartazio:
a cursory glance this looks fine to me, @dolio  @ekmett  @hvr  thoughts?

cartazio:
does this fix work with respect the issues in https://ghc.haskell.org/trac/ghc/ticket/9806 ?

also: this should have a test! 
(i'll be adding the first member of a regression suite subsection to the tests in my own PR this weekend, maybe that would be a reasonable analogous guide?)

sergv:
I'll add a tests for this, preferably after there will be a guide on how to do regression tests, i.e. after your PR.

Could you please clarify which issues from https://ghc.haskell.org/trac/ghc/ticket/9806 do you have in mind? It seems that that ticket talks mostly about system malloc and free functions, while in this PR everything boils down to `newAlignedPinnedByteArray` which bypasses system functions.

mgsloan:
Seems like a good change to me.

dolio:
Merging this. In the future, though. It'd be nice if functionality patches didn't come with random whitespace fixes.

cartazio:
@dolio  did we ever fix this proerly?

cartazio:
@dolio  did we ever add a test for this? @sergv  care to help? :)

sergv:
@cartazio I tested it locally in for AVX operations and it worked. I'll submit a test case.

cartazio:
Excellent! Do we have a unit test suite? Or do we need to creat one?

sergv:
@cartazio It seems that there's no unit test suite. It's not even among the dependencies - `test-framework-hunit` is not there. I think we need to create a new one.

cartazio:
Agreed! Let's make it a separate compilation unit.  If you wanted to use
hspec to write the suite instead of test framework that would probably fine
or just use test frameworks hunit.  As long as we put version on any deps
and its a stable pvp friendly dep alls good.

sergv:
@cartazio Bump for #131?

PR #101, opened 2015-11-02
Add examples for some functions in Data.Vector

MaxGabriel:
Here's an image of what the generated Haddocks look like:
image

I'm not sure if you have strong preferences on how exactly to format the code examples; I can be flexible with that.

### DIFF
--- a/Data/Vector.hs
+++ b/Data/Vector.hs
@@ -1245,79 +1245,256 @@ ifoldr' = G.ifoldr'
 -- -----------------
 
 -- | /O(n)/ Check if all elements satisfy the predicate.
+--
+-- ==== __Examples__
+--
+-- @
+-- __>>> V.'all' (== "hello") $ V.'replicate' 10 ("hello")__
+-- True
+-- @
+--
+-- @
+-- __>>> V.'all' (== "foo") $ V.'fromList' []__
+-- True
+-- @
 all :: (a -> Bool) -> Vector a -> Bool
 {-# INLINE all #-}
 all = G.all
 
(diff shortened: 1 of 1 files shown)


MaxGabriel:
It looks like the `git clone` failed for the [GHC=head build](https://travis-ci.org/haskell/vector/jobs/88704511):

```
75.13s$ git clone --depth=50 https://github.com/haskell/vector.git haskell/vector
Cloning into 'haskell/vector'...
fatal: unable to access 'https://github.com/haskell/vector.git/': Couldn't resolve host 'github.com'
The command "eval git clone --depth=50 https://github.com/haskell/vector.git haskell/vector" failed. Retrying, 2 of 3.
```

That may need to be retried.

cartazio:
this is a cool idea, @dolio  and i are doing some cleanup on master at the moment, but having doc tests style examples that are also part of our tests would be pretty sweet, pardon the slow turn around on this. would that be something you'd be interested in helping drive?

MaxGabriel:
I haven't been working on Haskell stuff for awhile, so realistically, probably not :(

Bodigrim:
This looks like a first step for #296, doesn't it?

Bodigrim:
This is not an acceptable format for `doctest`, is it?

5outh:
Can this be merged? Even if it's not doctest-compatible it's a step in the right direction of adding good examples.

Bodigrim:
We run doctests on CI, so this is pretty much a deal breaker. But a similar in scope PR with passing doctests (which is not hard at all) could be very well accepted.

Shimuuar:
Superceded by #363 and #364

PR #102, opened 2015-11-10
Add Hashable Vector instances

phadej:
Related to https://github.com/tibbe/hashable/issues/108#

### DIFF
--- a/Data/Vector.hs
+++ b/Data/Vector.hs
@@ -195,6 +195,8 @@ import qualified Data.Traversable as Traversable
 import Data.Monoid   ( Monoid(..) )
 #endif
 
+import Data.Hashable (Hashable(..))
+
 #if __GLASGOW_HASKELL__ >= 708
 import qualified GHC.Exts as Exts (IsList(..))
 #endif
@@ -351,6 +353,9 @@ instance Traversable.Traversable Vector where
 
   {-# INLINE sequence #-}
   sequence = sequence
+
+instance (Hashable a) => Hashable (Vector a) where
+  hashWithSalt salt = hashWithSalt salt . toList
 
 -- Length information
 -- ------------------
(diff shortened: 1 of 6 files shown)


cartazio:
I'm personally against this becsuse
A) hashable is pretty Broken
B) it won't be more efficient than calling hashable in from list.
C) it won't be the best possible hash function for a vector anyways

cartazio:
I mean to list.  I'm still waking up.  Pardon.

Also it will break any user land hashable orphans but not really solve any
problems.

I guess my main point is that hashable is a bad design and codifying it
further into more core libraries is something I'm not in favor of :)

phadej:
A) please explain
B) I cannot derive (using `Generic`) `Hashable` for types with `Vector` fields, because `Vector` isn't `Hashable`. I have to write boilerplate just for that.
C) `Hashable` is meant to provide hash function so values can be used as keys in hash-based data structures, `unordered-containers`. For that current function is good and fast enough. To clarify `Hashable` isn't meant to be good cryptographic hash.

> This package \[ `hashable` \] defines a class, Hashable, for types that can be converted to a hash value. This class exists for the benefit of hashing-based data structures.

Also I'm unaware of any libraries with hashable orphans for vectors. AFAIK there aren't one on Hackage. If there is one, I'd happy to use it, but I'd rather not to create any if there aren't.

andrewthad:
It would be nice to have `Hashable` instances for vectors. I used vectors a lot in several data types that have `Hashable` instances, and it's annoying not to be able to get this using generics. I don't understand the claim that "hashable is a bad design". It works fine for what it's supposed to do.

phadej:
@andrewthad use https://hackage.haskell.org/package/vector-instances-3.4/docs/Data-Vector-Instances.html

PR #104, opened 2015-12-11
Fix compilation with primitive master

thomie:


cartazio:
@thomie  this patch seems to fail on head and <=7.4

thomie:
Thanks. I fixed the compilation with <= 7.4. The current `#ifdef` is hopefully future proof.

Any patch will fail with GHC HEAD though, because @hvr increased the version number of base to 4.9, but there is no release of `primitive` yet that allows base 4.9.

Data/Vector.hs
```diff
@@ -161,7 +161,11 @@ module Data.Vector (
 
 import qualified Data.Vector.Generic as G
 import           Data.Vector.Mutable  ( MVector(..) )
+#if MIN_VERSION_primitive(0,6,2) || __GLASGOW_HASKELL__ >= 711
+import           Data.Primitive.Array hiding ( fromList, fromListN )
```
hvr: We can actually avoid the heavy CPP hammer...

thomie: Qualified import?

hvr: almost... qualified use =)

thomie:
Thanks.

PR #109, opened 2016-02-18
add Generics instance to Data.Vector

kuk0:
we want Generics for everything (otherwise we can't derive Generics for any data types which contain these without creating orphan instances)

the end-goal is to have all of these: https://gist.github.com/kuk0/92fff1c91c526ffb0a9e
and be able to derive instances for e.g. TextShow or Text.PrettyPrint.Generic.Pretty

### DIFF
--- a/Data/Vector.hs
+++ b/Data/Vector.hs
@@ -197,6 +197,7 @@ import Data.Monoid   ( Monoid(..) )
 
 #if __GLASGOW_HASKELL__ >= 708
 import qualified GHC.Exts as Exts (IsList(..))
+import GHC.Generics
 #endif
 
 
@@ -351,6 +352,30 @@ instance Traversable.Traversable Vector where
 
   {-# INLINE sequence #-}
   sequence = sequence
+
+type Rep1Vector = D1 D1Vector (C1 C1Vector (S1 NoSelector (Rec1 [])))
+
+instance Generic1 Vector where
+  type Rep1 Vector = Rep1Vector
+  from1 v = M1 (M1 (M1 (Rec1 $ toList v)))
+  to1 (M1 (M1 (M1 (Rec1 l)))) = fromList l
+
+data D1Vector
+data C1Vector
+
+instance Datatype D1Vector where
+  datatypeName _ = "Vector"
+  moduleName   _ = "Data.Vector"
+
+instance Constructor C1Vector  where
+  conName _ = "Vector.fromList"
+
+type Rep0Vector a = D1 D1Vector (C1 C1Vector (S1 NoSelector (Rec0 [a])))
+
+instance Generic (Vector a) where
+  type Rep (Vector a) = Rep0Vector a
+  from v = M1 (M1 (M1 (K1 $ toList v)))
+  to (M1 (M1 (M1 (K1 l)))) = fromList l
 
 -- Length information
 -- ------------------


cartazio:
possibly silly question: why can't we do DeriveGeneric directly on the various Vector data types?

Shimuuar:
DeriveGenerics would break data abstraction and  conversion to/from list is likely to break performance

kuk0:
@cartazio: not a silly question; deriving the Generic instance is possible, however, i'm not sure how useful it is; internally, Vector is made of an Int and a primitive Array; if i want to derive any other instance based on the Generic instance, i would (imho) have to implement it manually because 1.) i want it to behave as if it was a list, not an Int + Array, and 2.) Array probably doesn't have the required instance anyway - this is at least true for my motivating type classes TextShow and Pretty
(also dealing with primitive Arrays will be messy)

@Shimuuar: ad "conversion to/from list is likely to break performance" - not necessarily - you can still define your own instance which is not based on Generics; (i'm actually trying that with TextShow for Vector and i might be doing sth. wrong, but so far my simple/naive attempts at writing something faster than just the derived TextShow instance based on lists failed)

cartazio:
I'm uncomfortable with the behave like a list rather than a sized array
desugaring.  Part of the value of unboxed / storable structures is their
very memory layout, and your design choice seems to preclude that.  Maybe
I'm not understanding your point.

I thought the generics instances for vector would / should back out to the
data structures in the promitive package, and we can certainly make sure
the right infrastructure is available there.

kuk0:
@cartazio: The point of Generics is representing data types via sums and products of simpler types. I can certainly just DeriveGenerics and then go write my TextShow instance as `showb = showb . Vector.toList`.
My point is:

1. the instance we get from DeriveGenerics (imho) unnecessarily exposes the internal implementation (offset, length, the array that contains the vector)
2. the derived Generic instance is totally useless (imho).

Let's talk about 2.)
Afaik, Vector is implemented as some segment inside of a (potentially larger) Array, i.e., the Array can be: [some garbage] my vector [some garbage], and the offset and length point to my vector.
I can't imagine an instance which could be simply derived from this Generic instance. For example:

* TextShow:  i want to show just my vector part, not the whole array + offset
* Prettyprint: i want to show just my vector part, not the whole array + offset 
* Serialize: i want to serialize just my vector, not the whole array + offset
* ToJSON: .... same....

Can you think of a usecase where we would want to use the internal representation of the Vector?
After all, even the existing Show instance for Vector shows it as `fromList [...]` - it doesn't show the internal representation.

If i represent the Vector by a list in the Generic instance, i can get all the other instances for free:
```
instance TextShow a => TextShow (Vector a) where
  showbPrec = genericShowbPrec
```
If this is too slow for me, i can still write my own optimized version of `TextShow (Vector a)`, but at least i have a good, working default for free.
If we derive the Generic instance, we will have to implement all the other instances from scratch.

RyanGlScott:
I think you're trying to stretch GHC generics beyond its intended purpose. GHC generics is intended to be used with _regular_ datatypes—that is, datatypes with a predictable, algebraically defined structure. GHC generics necessarily exposes the underlying structure in order to be able to perform generic operations on it.

`Vector` and friends are definitely not what I'd consider to be "regular". The internals of `Vector` are quite grimy compared to the abstract, list-esque interface it provides, and users are almost never expected to rely on the internal structure of a `Vector`.

Deriving a `Generic` instance for `Vector`s would certainly be bad. And hand-crafting a `Generic` instance that lies to you by pretending that `Vector` is actually a list in disguise is even worse! There's a reason you can't write your own `Generic` instances with `-XSafe` on.

I think you're trying to solve your problem in the wrong way. There is a `text-show-instances` package that provides `TextShow` instances for the `vector` datatypes. Not sure about `Text.PrettyPrint.Generic.Pretty`.

cartazio:
well said @RyanGlScott, this articulates *why* i'm opposed hand written generics for vector.

likewise, [GHC.Generics](https://hackage.haskell.org/package/base-4.8.2.0/docs/GHC-Generics.html) (as ryan indirectly states) CANNOT express the correct data structure (with the right asymptotic complexity), whereas it can for ordinary algebraic data types.  To support a generics for Vector correctly, we'd need to have GHC.Generics to add support for boxed homogenous vectors which is a pretty big ask, and likewise would have to talk about generics support for Data.Primitive.Array (@dolio :) ) and Data.Array in the arrays package :)

kuk0:
ok

PR #123, opened 2016-07-24
Use mapM in createT because ST wasn't Applicative in 7.0

dolio:


cartazio:
@dolio Q: does this sensibly have an analogous specialized type for the Unboxed,Boxed and Storable modules?

dolio:
This already is in those modules. They're just copies of the generic API like most other things.

PR #125, opened 2016-07-24
Fix iscanr/l functions to fuse with correct lengths

dolio:
They were using inplace with id, while scans increase the number of elements by
1. iscanr functions are incorrect, though.

### DIFF
--- a/Data/Vector/Generic.hs
+++ b/Data/Vector/Generic.hs
@@ -1822,18 +1822,22 @@ scanr' f z = unstreamR . Bundle.scanl' (flip f) z . streamR
 -- | /O(n)/ Right-to-left scan over a vector with its index
 iscanr :: (Vector v a, Vector v b) => (Int -> a -> b -> b) -> b -> v a -> v b
 {-# INLINE iscanr #-}
-iscanr f z =
+iscanr f z v =
     unstreamR
-  . inplace (S.scanl (flip $ uncurry f) z . S.indexed) (+1)
+  . inplace (S.scanl (flip $ uncurry f) z . S.indexedR n) (+1)
   . streamR
+  $ v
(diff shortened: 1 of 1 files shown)


Data/Vector/Generic.hs
```diff
@@ -1822,18 +1822,22 @@ scanr' f z = unstreamR . Bundle.scanl' (flip f) z . streamR
 -- | /O(n)/ Right-to-left scan over a vector with its index
 iscanr :: (Vector v a, Vector v b) => (Int -> a -> b -> b) -> b -> v a -> v b
 {-# INLINE iscanr #-}
-iscanr f z =
+iscanr f z v =
     unstreamR
```
cartazio: huh, i didnt realize our Stream tooling could do "from the Right", cool!

dolio: I think that's new with the bundle stuff. Unfortunately this function probably breaks fusion, because of the need for length information.

cartazio: I thought vector stream fusion can support an exact size? Using the Exact
size constructor

Hrmmmm. Maybe this code should use that?

dolio: So, the issue is, you don't necessarily have an exact size.

If you have an actual vector, you have an exact size. But then you haven't fused. And there are fusion rules that eliminate stuff like `streamR (unstreamR bundle) = bundle` (not quite that), and not all bundles have an exact size. We could create a version of `inplace` that gives the bundle's `Size`, but if it's `Max` or `Unknown`, we don't know what to do for the indices; we'd just have to compute the length of the `Stream`.

A better option might be to write a fully fleshed out `iscanr` for `Bundle` and use it.

PR #129, opened 2016-07-25
added MonadZip Vector instance and base version guard

fhaust:
Adding `MonadZip` instances for `Vector` types allows to use `-XMonadComprehensions` and `-XParallelComprehensions` together:

    >>> :set -XMonadComprehensions
    >>> :set -XParallelListComp
    >>> import qualified Data.Vector as V
    >>> let as = V.fromList [0..10]
    >>> let bs = V.fromList [0..10]
    >>> [ a + b | a <- as | b <- bs ]
    fromList [0,2,4,6,8,10,12,14,16,18,20]

[previous pull request](https://github.com/haskell/vector/pull/78#issuecomment-234935758)

### DIFF
--- a/Data/Vector.hs
+++ b/Data/Vector.hs
@@ -168,6 +168,7 @@ import Control.DeepSeq ( NFData, rnf )
 import Control.Monad ( MonadPlus(..), liftM, ap )
 import Control.Monad.ST ( ST )
 import Control.Monad.Primitive
+import Control.Monad.Zip
 
 import Prelude hiding ( length, null,
                         replicate, (++), concat,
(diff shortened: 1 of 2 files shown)


fhaust:
Ok ... these should work now. `MonadZip` was not introduced before `base-4.4.0` so I had to sprinkle CPP conditionals everywhere ... is that acceptable?

cartazio:
I think @dolio and I agreed that vector master and thus the 0.12
release needn't support older than ghc 7.4 , which means you needed use
that cpp and instead should put a base >= 4.4.0  bound in he cabal file. We
may ultimately decide to have a more aggressive lower bound than that.  But
for now I think there's clear agreement on that front

Let's let @dolio chime in too, but subject to his agreement (plus suitable
code review )

A) remove the cpp :)

B) add the bound in vector.cabal

C) remove the pre 7.4 ghc configs for the Travis file

Then we shall cut ribbon and merge the good stuff in

cartazio:
Thst you do not need that min base 4.4 cpp.  Woops. Typo.

dolio:
Yeah, apologies about the version confusion. I'll just merge this, and then fix up the CPP when I go through and increase the min bounds (maybe in conjunction with fixing up the constraints patch).

PR #130, opened 2016-07-25
Add uniq function

### LINKED ISSUE #107: Feature request for uniq
It would be nice to have `uniq` for operating on vectors. It would have the following type signature:

    uniq :: Eq a => Vector a -> Vector a

and it would remove any adjacent identical elements (like the linux program does). I'm just going to write this myself for a project I'm working on, but it would be nice to have one that I could trust to be efficient (since I'm not particularly good with optimizing haskell / reading core).

Shimuuar:
It removes duplicate adjacent elenmts from vector. Should fix #107 

No tests so far. What is current story about test suite?

### DIFF
--- a/Data/Vector.hs
+++ b/Data/Vector.hs
@@ -113,7 +113,7 @@ module Data.Vector (
   -- * Working with predicates
 
   -- ** Filtering
-  filter, ifilter,
+  filter, ifilter, uniq,
   mapMaybe, imapMaybe,
   filterM,
   takeWhile, dropWhile,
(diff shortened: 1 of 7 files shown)


cartazio:
Great, we will take a look at it shortly

dolio:
The test suite isn't wonderful, but it does work (except on 7.10). And I made it actually run on travis recently.

This is close to the old run-length-encoding trick for lists, no? So it could be tested against `map head . group`?

Shimuuar:
I've added test and rebased to current master.

I simply vaguely remembered some talk about upgrade/reorganization of test suite. Also tests do work on GHC 7.10. They just take forever and 6GB of memory to build.

dolio:
Looks good.

And yeah, I can actually build the tests relatively quickly on my desktop with 7.10, but I think it chews up around 10 GB of memory in the process. Sadly, travis doesn't give you such resources.

cartazio:
@dolio should we add a bikeshed ticket for the 0.12 checklist to futz with name choice ?

dolio:
Sure.

PR #134, opened 2016-08-16
Correct semigroups conditional version bounds

RyanGlScott:
`Data.Semigroup` was introduced in `base-4.9` (GHC 8.0), but `impl` flag was causing `vector` to depend on the `semigroups` library even when GHC 8.0 was being used. This fixes it so that `vector` doesn't depend on `semigroups` for GHC 8.0.

### DIFF
--- a/vector.cabal
+++ b/vector.cabal
@@ -155,7 +155,7 @@ Library
                , primitive >= 0.5.0.1 && < 0.7
                , ghc-prim >= 0.2 && < 0.6
                , deepseq >= 1.1 && < 1.5
-  if !impl(ghc > 8.0)
+  if impl(ghc < 8.0)
     Build-Depends: semigroups >= 0.18 && < 0.19
 
   Ghc-Options: -O2 -Wall


hvr:
@RyanGlScott 

I don't agree with this commit because I can't reproduce the problem this is supposed to fix, and also because

    if !impl(ghc > 8.0)

means, "if GHC is **not** used, or GHC is used and its version is less than or equal to 8.0"

however, there is no official GHC that has the version number "8.0"; it's always a 3-part version, i.e. "8.0.x". 

whereas 

    if impl(ghc < 8.0)

means, "if GHC is used, and its version is smaller than 8.0"

RyanGlScott:
@hvr, I apologize, it appears I didn't understand how `!` interacts with `impl`. I'll revert this commit.

That being said, that behavior is quite unintuitive to me—I would expect `if !impl(ghc > 8.0)` to be logically equivalent to `impl(ghc <= 8.0)`, and indeed the [`Cabal` documentation](https://github.com/haskell/cabal/blob/9b4712583ca390d0f034cc66eca4de51c00cc8d0/Cabal/doc/developing-packages.markdown#L1717) doesn't seem to mention anything about this.

Your point about `makeVersion [8,0]` being less than `makeVersion [8,0,x]` is something I didn't consider, admittedly.

cartazio:
You can write intersections of upper and lower bounds of impl versions , or
at least cabal accepts them :)

RyanGlScott:
Right, that's what's interesting about `!`'s interaction with `impl`. At first glance, I would expect something like `if !impl(ghc >= 7.10 && < 8.0)` to be exactly the same as `impl(ghc < 7.10 || ghc >= 8.0)`, de Morgan–style. But that's not completely true, apparently—there's an additional stipulation that "If GHC isn't used, __or__ `impl(ghc < 7.10 || ghc >= 8.0)`". That part wasn't clear at all to me, since I've never seen that written down anywhere.

cartazio:
Oh.  I do the and outside the impl()

hvr:
@RyanGlScott it's not so obvious, and it may be a good idea to write that one down. 

IMO:  `impl(ghc >= x.y)` is trivially equivalent to `impl(ghc) && impl(ghc >= x.y)`, and if you negate that, you end up with `!impl(ghc) || !impl(ghc >= x.y)`

PR #143, opened 2016-11-22
add initial groupby

codygman:
I'm guessing more will need to be done to get this merged, but I think it would be pretty useful.

### DIFF
--- a/Data/Vector/Generic.hs
+++ b/Data/Vector/Generic.hs
@@ -104,6 +104,10 @@ module Data.Vector.Generic (
   -- ** Partitioning
   partition, unstablePartition, span, break,
 
+  -- ** Grouping
+  groupBy
+
+
   -- ** Searching
   elem, notElem, find, findIndex, findIndices, elemIndex, elemIndices,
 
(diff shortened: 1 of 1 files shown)


Data/Vector/Generic.hs
```diff
@@ -2179,3 +2183,24 @@ dataCast :: (Vector v a, Data a, Typeable1 v, Typeable1 t)
          => (forall d. Data  d => c (t d)) -> Maybe  (c (v a))
 {-# INLINE dataCast #-}
 dataCast f = gcast1 f
+
+{-# INLINE groupBy #-}
+groupBy k xs =
+  switchL []
```
Bodigrim, line 2189: What's the point? Why not pattern-match immediately?
```haskell
groupBy k [] = [] 
groupBy k (h : t) = ...
```

Data/Vector/Generic.hs
```diff
@@ -2179,3 +2183,24 @@ dataCast :: (Vector v a, Data a, Typeable1 v, Typeable1 t)
...
+  (\h t ->
+      let n = 1 + findIndexOrEnd (not . k h) t
+      in  unsafeTake n xs : groupBy k (unsafeDrop n xs))
+  xs
+
+-- | 'findIndexOrEnd' is a variant of findIndex, that returns the length
+-- of the string if no element is found, rather than Nothing.
```
Bodigrim, line 2196: Is it faster than `fromMaybe len (findIndex p)`?
I think it is better to pattern-match on result of `findIndex` in `groupBy`: if it's `Nothing` you do not need to make another recursive call.

Shimuuar:
Superseded by #427

PR #153, opened 2017-02-12
Include modules that test-suites depend on in other-modules

### LINKED ISSUE #138: Please add support for QuickCheck 2.9.x
The test suite constrains `QuickCheck ==2.7.*`, but that restrictions seems unnecessary, i.e. the build works fine if that version range is extended to incude 2.8.\* (and possibly 2.9.x as well, though I didn't test that).

RyanGlScott:
As noted in https://github.com/haskell/vector/issues/138#issuecomment-279217723, `vector`'s test suites do not compile when obtained from Hackage. That is because a module that the test suites depend on (`Tests.Vector.UnitTests`) is not included with the tarball that gets uploaded to Hackage, so when one attempts to compile the test suites from Hackage, you get a "Failed to load interface" error.

This fixes the issue by adding `Tests.Vector.UnitTests` (and other dependent modules) to an `other-modules` stanza in the test suites, which makes `Cabal` include them in a release tarball. As a result, there's no need to list them explicitly in `extra-source-files` anymore.

Since this issue is keeping `vector` [partially blocked from Stackage](https://github.com/fpco/stackage/blob/ebcc6780ddde70a860ecb5667a5a5f690def3d8a/build-constraints.yaml#L2997), it'd probably be a good idea to make a minor release with this change.

Fixes #138.

### DIFF
--- a/vector.cabal
+++ b/vector.cabal
@@ -50,11 +50,6 @@ Extra-Source-Files:
       tests/LICENSE
       tests/Setup.hs
       tests/Main.hs
-      tests/Boilerplater.hs
-      tests/Utilities.hs
-      tests/Tests/Move.hs
-      tests/Tests/Bundle.hs
-      tests/Tests/Vector.hs
       benchmarks/vector-benchmarks.cabal
       benchmarks/LICENSE
       benchmarks/Setup.hs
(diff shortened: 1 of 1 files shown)


cartazio:
thanks! :)

cartazio:
@RyanGlScott it looks like @hvr  and @dolio  are the two folks who are active that have uploader permissions (bos and romanL of ancient times )

PR #155, opened 2017-02-17
Make sure that 'length' can be inlined (Fixes #97)

### LINKED ISSUE #97: Inefficient code generated for length
Here is a a function that checks if two unboxed vectors have the same length:

``` haskell
module Test where

import qualified Data.Vector.Unboxed as V

sameLength :: V.Vector Int -> V.Vector Int -> Bool
sameLength x y = V.length x == V.length y
```

I would expect `sameLength` to compile into code that just pattern matches on its arguments. Unfortunately, what I get is the following Core:

```
sameLength :: V.Vector Int -> V.Vector Int -> Bool
sameLength =
  \ (x_a2P6 :: V.Vector Int) (y_a2P7 :: V.Vector Int) ->
    case Test.$slength x_a2P6 of _ { GHC.Types.I# x1_a48j ->
    case Test.$slength y_a2P7 of _ { GHC.Types.I# y1_a48n ->
    GHC.Prim.tagToEnum# @ Bool (GHC.Prim.==# x1_a48j y1_a48n)
    }
    }

Test.$slength [InlPrag=[NEVER]] :: V.Vector Int -> Int
[GblId, Arity=1, Caf=NoCafRefs, Str=DmdType <S,1*U(A,U,A)>m]
Test.$slength = ...
```

For some reason, the `$slength` function, which computes the length of the vector, is marked with `InlPrag=[NEVER]`. Hence the `sameLength` function does lots of unnecessary boxing and unboxing.

In reality, this boxing and unboxing happens inside a tight loop where it causes a ~2x slowdown in my code.

takano-akio:
Previously there was a cycle in the dependency graph of functions,
consisting of {stream, clone, length, unsafeCopy}. This was causing GHC
to mark one of these functions, length, as a loop breaker.

This commit breaks this down this strongly-connected component by
removing the edge from length to stream.

I have confirmed that the test case in #97 is compiled into nice code with this change.

### DIFF
--- a/Data/Vector/Generic.hs
+++ b/Data/Vector/Generic.hs
@@ -228,7 +228,7 @@ import qualified Data.Traversable as T (Traversable(mapM))
 -- | /O(1)/ Yield the length of the vector
 length :: Vector v a => v a -> Int
 {-# INLINE length #-}
-length = Bundle.length . stream
+length = Bundle.length . stream'
 
 -- | /O(1)/ Test whether a vector is empty
 null :: Vector v a => v a -> Bool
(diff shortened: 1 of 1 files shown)


cartazio:
@takano-akio cool!

this raises an interesting question: how should we track making sure all these little changes  stay the correct ones?

@takano-akio @dolio @hvr should we figure out having some sort of benchmark suite integrated into the project? (i feel like we'd need to use something like cabal.project / cabal-new build style tricks to be able to use criterion or friends correctly for this). But we do want to track  which of equivalent implementations of different pieces perform better in a given version of vector and ghc. (cause measuring is the only way!)

dolio:
I've done a little experimenting (in my local copy of vector-algorithms), and you can definitely use cabal.project to do a criterion test suite even though it depends on vector.

PR #161, opened 2017-04-05
Make it build with ghc 8.2

erikd:
Requires patched [test-framework](https://github.com/erikd/test-framework).

### DIFF
--- a/vector.cabal
+++ b/vector.cabal
@@ -144,7 +144,7 @@ Library
   Install-Includes:
         vector.h
 
-  Build-Depends: base >= 4.5 && < 4.10
+  Build-Depends: base >= 4.5 && < 4.11
                , primitive >= 0.5.0.1 && < 0.7
                , ghc-prim >= 0.2 && < 0.6
                , deepseq >= 1.1 && < 1.5


RyanGlScott:
Thanks. However, I'm reluctant to merge this at the moment, since the `-O2` test suite fails to build on Travis due to an [out-of-memory error](https://travis-ci.org/haskell/vector/jobs/218749281#L1270). On my local machine with 16 GB of memory, I can at least build that test suite (after waiting quite a while), but then the test suite appears to run forever.

RyanGlScott:
I've filed a corresponding GHC issue here: [Trac #13535](https://ghc.haskell.org/trac/ghc/ticket/13535)

cartazio:
Is this because of the bundle rather than stream rep or because of the
weird way the vector test suite is written. Or because of fusio. Blowup ?


Happy to hear O2 is tickling some issues

phadej:
Without this testing GHC-8.2.1-rc1 is rather difficult

dolio:
Ran tests locally (only took 20GB of memory to compile). They passed. So I'm merging and bumping the dependencies on hackage.

PR #168, opened 2017-05-01
No more `error: division by zero` when trying to print or operate on empty storable vectors

mihaimaruseac:


dolio:
Looks good. At first I was thinking we could do better, because you need no storage for `()` and presumably anything else of size 0. But then I realized that `mallocVector` will take care of this as well as can be done, so it's fine.

cartazio:
@dolio : 
reasoning  being that the code is now 

```

  basicUnsafeNew n
    | n < 0 = error $ "Storable.basicUnsafeNew: negative length: " ++ show n
    | n > mx = error $ "Storable.basicUnsafeNew: length too large: " ++ show n
    | otherwise = unsafePrimToPrim $ do
        fp <- mallocVector n
        return $ MVector n fp
    where
      size = sizeOf (undefined :: a) `max` 1
      mx = maxBound `quot` size :: Int
```
and mallocVector uses the storable width times the number of elements to determine size of allocation?

dolio:
Yeah `mallocVector` will allocate 0 bytes for a size 0 type. I suppose in principle we could do something a little dicey, like put a null pointer in there, and trust that the `peek`/`poke` in the `Storable` instance don't do anything with the pointer. But that's a pretty marginal win.

Issue #172, opened 2017-06-05
Is there an efficient way to `break`/`span`/`findIndex` from the right end?


leftaroundabout:
Unlike with lists, for arrays these operations make quite as much sense from both ends of the data structure, yet `vector` only offers them from the left, i.e. searching over ascending indices.

Is it possible to write an efficient<sup>†</sup> high-level implementation of the _descending_ version, in terms of the already existing functions? Else I would like to request a right-starting version at least of `findIndex`, which would readily allow defining the others too.

<hr>

<sup>†</sup><sub>By efficient I mean, comparable to a C implementation that runs an interrupted loop over a descending array-index. It should definitely be only _O_ (_n_ ­– _k_), not _O_ (_n_) or _O_ (_k_) where _n_ is the number of elements and _k_ the rightmost matching index.</sub>

leftaroundabout:
It definitely is possible with `foldl`:
```haskell
findIndexR :: (a->Bool) -> V.Vector a -> Maybe Int
findIndexR pred v = (V.length v-)
  <$> foldl (\a x -> if pred x
                      then Just 1
                      else succ<$>a) Nothing v
```
but I reckon that's a lot slower than the `Bundle` based left-starting version.

cartazio:
if you look at the underlying implementation in https://github.com/haskell/vector/blob/master/Data/Vector/Fusion/Stream/Monadic.hs#L852

namely 
```haskell
-- | Yield 'Just' the index of the first element that satisfies the monadic
-- predicate or 'Nothing' if no such element exists.
findIndexM :: Monad m => (a -> m Bool) -> Stream m a -> m (Maybe Int)
{-# INLINE_FUSED findIndexM #-}
findIndexM f (Stream step t) = findIndex_loop SPEC t 0
  where
    findIndex_loop !_ s i
      = do
          r <- step s
          case r of
            Yield x s' -> do
                            b <- f x
                            if b then return $ Just i
                                 else findIndex_loop SPEC s' (i+1)
            Skip    s' -> findIndex_loop SPEC s' i
            Done       -> return Nothing
```
you'll see that the from the left version is tied deeply to the stream fusion representation.

a search from the right will not play well with the fusion framework as such, or at least i have trouble seeing how it would work this evening. perhaps @dolio  has thoughts on this.

point being: its a valid combinator to want, but its also quite easy to quickly write in user space... and theres definitely the subtle issue of making sure it'd play nicely with the optimization / fusion expectations users have for vector. 

[...]

PR #174, opened 2017-06-10
A version of `findIndex` for the last matching element.

leftaroundabout:
Discussed in

https://github.com/haskell/vector/issues/172

### DIFF
--- a/Data/Vector/Generic.hs
+++ b/Data/Vector/Generic.hs
@@ -106,7 +106,7 @@ module Data.Vector.Generic (
   partition, partitionWith, unstablePartition, span, break,
 
   -- ** Searching
-  elem, notElem, find, findIndex, findIndices, elemIndex, elemIndices,
+  elem, notElem, find, findIndex, findIndexR, findIndices, elemIndex, elemIndices,
 
   -- * Folding
   foldl, foldl1, foldl', foldl1', foldr, foldr1, foldr', foldr1',
@@ -1499,6 +1499,12 @@ find f = Bundle.find f . stream
 findIndex :: Vector v a => (a -> Bool) -> v a -> Maybe Int
 {-# INLINE findIndex #-}
 findIndex f = Bundle.findIndex f . stream
+
+-- | /O(n)/ Yield 'Just' the index of the /last/ element matching the predicate
+-- or 'Nothing' if no such element exists.
+findIndexR :: Vector v a => (a -> Bool) -> v a -> Maybe Int
+{-# INLINE findIndexR #-}
+findIndexR f v = fmap (length v - 1 -) . Bundle.findIndex f $ streamR v
 
 -- | /O(n)/ Yield the indices of elements satisfying the predicate in ascending
 -- order.
--- a/benchmarks/Algo/FindIndexR.hs
+++ b/benchmarks/Algo/FindIndexR.hs
@@ -1,0 +1,24 @@
+module Algo.FindIndexR (findIndexR, findIndexR_naive, findIndexR_manual)
+where
+
+import Data.Vector.Unboxed (Vector)
+import qualified Data.Vector.Generic as V
+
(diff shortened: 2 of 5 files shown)


Data/Vector/Generic.hs
```diff
@@ -1455,6 +1455,12 @@ findIndex :: Vector v a => (a -> Bool) -> v a -> Maybe Int
 {-# INLINE findIndex #-}
 findIndex f = Bundle.findIndex f . stream
 
+-- | /O(n)/ Yield 'Just' the index of the /last/ element matching the predicate
+-- or 'Nothing' if no such element exists.
+findIndexR :: Vector v a => (a -> Bool) -> v a -> Maybe Int
+{-# INLINE findIndexR #-}
```
leftaroundabout: I'm actually not sure if it makes sense to inline this. [dolio remarked](https://github.com/haskell/vector/issues/172) that this function won't be able to use any stream fusion. I know too little about vector's optimisations to judge this.

cartazio:
1) could you also add a benchmark comparing this vs explicitly looping over the vector too? (we've established mostly that this isn't going to fuse with the current design, not how to make it most performant)

2) should the name be `findLastIndex` rather than `findIndexR` ?  
It seems to me that because streams and vector indices are "oriented", 'findLastIndex' might be better :)

leftaroundabout:
1. Yeah, I can add such a benchmark. Do you mean, I should add two new function checks to `benchmarks/Main.hs`' `defaultMain` list?

2. I don't have much of an opinion on the naming. `findLastIndex` is certainly nice descriptive. But `findIndexR` should IMO be clear enough too. And it's consistent with [`Data.Sequence.findIndexR`](http://hackage.haskell.org/package/containers-0.5.10.2/docs/Data-Sequence.html#v:findIndexR) as well `streamR` on which it is based.

leftaroundabout:
The benchmark suggests that this function is indeed worthwhile, in that it's 40% faster than a manual loop (it's possible that I've missed some optimisation that should be used in the loop, though. But then, that could also happen to other people, so...)
```
benchmarking findIndexR
time                 592.0 μs   (589.8 μs .. 594.7 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 591.3 μs   (589.8 μs .. 593.0 μs)
std dev              5.563 μs   (4.182 μs .. 7.344 μs)

benchmarking findIndexR_naïve
time                 15.87 ms   (15.66 ms .. 16.08 ms)
                     0.999 R²   (0.998 R² .. 1.000 R²)
mean                 16.01 ms   (15.92 ms .. 16.17 ms)
std dev              316.9 μs   (207.2 μs .. 490.3 μs)

benchmarking findIndexR_manual
time                 977.0 μs   (968.0 μs .. 987.1 μs)
                     0.999 R²   (0.999 R² .. 1.000 R²)
mean                 978.0 μs   (973.1 μs .. 986.6 μs)
std dev              21.65 μs   (13.65 μs .. 34.29 μs)
```

benchmarks/Algo/FindIndexR.hs
```diff
@@ -0,0 +1,24 @@
...
+where
+
+import Data.Vector.Unboxed (Vector)
+import qualified Data.Vector.Generic as V
+
+findIndexR :: (Double -> Bool, Vector Double) -> Maybe Int
+{-# NOINLINE findIndexR #-}
+findIndexR = uncurry V.findIndexR
+
+findIndexR_naïve :: (Double -> Bool, Vector Double) -> Maybe Int
```
Bodigrim, line 11: Let's avoid Unicode in function names.

leftaroundabout:
@Bodigrim done

lehins:
This looks good to me. @Bodigrim do you have any more suggestions? The ones you've pointed out have already been taken care of. I'll merge the PR in a few days if I don't hear form you.

PR #179, opened 2017-07-05
Add MonadFix instance

### LINKED ISSUE #178: Add MonadFix instance for Vector
I *believe* this should be equivalent to the `MonadFix` instance for `[]`:

```haskell
instance MonadFix Vector where
  mfix f
    | V.null v0 = V.empty
    | otherwise = v
    where
      v = V.generate (V.length v0)
                     (\i -> if i == 0
                            then V.head v0
                            else f (v ! i) ! i)
      v0 = fix (f . V.head)
```

Of course, that `generate` can be replaced with something slightly more efficient, but I don't know if it's worth the trouble.

treeowl:
I *believe* this is equivalent to the instance for `[]`. Writing
QuickCheck properties for `mfix` seems pretty tricky, so I just
added a small unit test.

Fixes #178

### DIFF
--- a/Data/Vector.hs
+++ b/Data/Vector.hs
@@ -171,11 +171,12 @@ import qualified Data.Vector.Fusion.Bundle as Bundle
 
 import Control.DeepSeq ( NFData, rnf )
 import Control.Monad ( MonadPlus(..), liftM, ap )
-import Control.Monad.ST ( ST )
+import Control.Monad.ST ( ST, runST )
 import Control.Monad.Primitive
 
-
+import Control.Monad.Fix ( MonadFix (mfix) )
 import Control.Monad.Zip
+import Data.Function ( fix )
 
 import Prelude hiding ( length, null,
                         replicate, (++), concat,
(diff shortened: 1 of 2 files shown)


treeowl:
This implementation doesn't make any attempt to be fusion-friendly. That would seem to require digging into some internals. Unfortunately, I tend to get rather lost in `Bundle`.

dolio:
I don't think being fused is a realistic goal for this.

I'll try to look it over soon.

treeowl:
In case this helps: I just tested this implementation against the two list
monad examples in Levent Erkök's thesis, and it seems to check out.

dolio:
Do the properties of `mfix` here mean that boostrapping with `f ⊥` is just as good as `fix (head . f)` (because we just want to know the 'statically determined' length)? And then the `generate` is less complicated?

treeowl:
I believe so. The only trouble is that we end up with an extra call to the
passed function. I don't *think* we save much complexity over a slightly
simplified version. I'll post one or two options later if I have time.
Please poke me if I haven't done so by Monday.

On Jul 15, 2017 11:47 AM, "dolio" <notifications@github.com> wrote:

Do the properties of mfix here mean that boostrapping with f ⊥ is just as
good as fix (head . f) (because we just want to know the 'statically
determined' length)? And then the generate is less complicated?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/haskell/vector/pull/179#issuecomment-315542759>, or mute
the thread

Shimuuar:
Superseded by #312

PR #184, opened 2017-09-26
Make use of GHC call-stack simulation for the bounds-checked partial functions

leftaroundabout:
Until now, Vector's `(!)`, `head` etc., like other partial functions in Haskell, simply crashed the program without giving useful information as to where the problem originated – unless compiling with `-fprof`, which gives stack traces for everything but is horrible for performance.

GHC-8 has  added the [`HasCallStack`](https://hackage.haskell.org/package/base-4.9.1.0/docs/GHC-Stack.html#t:HasCallStack) constraint, which allows such functions to give meaningful error diagnosis without requiring global profiling. I enabled this on this library's checked-partial functions.

### DIFF
--- a/include/stacktracetools.h
+++ b/include/stacktracetools.h
@@ -1,0 +1,6 @@
+#if MIN_VERSION_base(4,9,0)
+import GHC.Stack (HasCallStack)
+#define CHECK(f) (withFrozenCallStack Ck.f __FILE__ __LINE__)
+#else
+#define HasCallStack (Eq ())
+#endif
--- a/Data/Vector.hs
+++ b/Data/Vector.hs
@@ -213,6 +213,9 @@ import qualified GHC.Exts as Exts (IsList(..))
 #endif
 
 
+#include "stacktracetools.h"
+
+
 -- | Boxed vectors, supporting efficient slicing.
 data Vector a = Vector {-# UNPACK #-} !Int
                        {-# UNPACK #-} !Int
@@ -450,7 +453,7 @@ null = G.null
 -- --------
 
 -- | O(1) Indexing
-(!) :: Vector a -> Int -> a
+(!) :: HasCallStack => Vector a -> Int -> a
 {-# INLINE (!) #-}
 (!) = (G.!)
 
@@ -460,12 +463,12 @@ null = G.null
 (!?) = (G.!?)
 
 -- | /O(1)/ First element
-head :: Vector a -> a
+head :: HasCallStack => Vector a -> a
 {-# INLINE head #-}
 head = G.head
 
 -- | /O(1)/ Last element
-last :: Vector a -> a
+last :: HasCallStack => Vector a -> a
 {-# INLINE last #-}
 last = G.last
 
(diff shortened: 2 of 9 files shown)


cartazio:
It would be good to make sure that this doesn’t interact poorly with ghc
fusion or optimization’s.

Likewise : with cabal new build profiled builds or debug builds are cached
in a pretty nice way.


That said : this could Be great: let’s make sure that it doesn’t have any
cost :)

include/stacktracetools.h
```diff
@@ -0,0 +1,6 @@
+#if MIN_VERSION_base(4,9,0)
+import GHC.Stack (HasCallStack)
+#define CHECK(f) (withFrozenCallStack Ck.f __FILE__ __LINE__)
+#else
+#define HasCallStack (Eq ())
```
leftaroundabout: That `Eq ()` fallback constraint was just a rough hack, to replace `HasCallStack` with something as simple as possible on GHC versions that don't support it yet. It turns out that hack doesn't work:
```
Data/Vector/Internal/Check.hs:122:15:
    Non type-variable argument in the constraint: Eq ()
    (Use FlexibleContexts to permit this)
    In the type signature for ‘checkIndex’:
      checkIndex :: (Eq ()) =>
                    String -> Int -> Checks -> String -> Int -> Int -> a -> a
```
Any ideas how this could be properly done without cluttering every signature with an `#if MIN_VERSION_base`?

leftaroundabout:
Sure, I agree. My idea was to only add `HasCallStack` to functions which are anyway “slow”, e.g. to `(!)` but not to `unsafeIndex`, so that any possible performance cost won't take action in spots that are deemed crucial for performance. But you're right of course that it should be checked that there are no unforeseen performance ramifications, I just don't know how how to go about this... I suppose it would be best to benchmark a whole sizable application before-and-after these changes?

include/stacktracetools.h
RyanGlScott: I would do what the [`call-stack` library does](http://hackage.haskell.org/package/call-stack-0.1.0/docs/src/Data-CallStack.html):

```haskell
#if MIN_VERSION_base(4,9,0)
import           GHC.Stack (HasCallStack)
#elif MIN_VERSION_base(4,8,1)
type HasCallStack = (?callStack :: GHC.CallStack)
#else
import GHC.Exts (Constraint)
type HasCallStack = (() :: Constraint)
#endif
```

This requires `base-4.5` (GHC 7.4) or later, but those are precisely the version bounds that `vector` has. Moreover, this shouldn't require the use of `FlexibleContexts`.

bgamari:
I suspect that as far as performance is concerned the callstacks generally shouldn't be a problem so long as the function is inlined. In this case I would expect that the callstack allocation will be floated in to the failure branch if one is necessary at all. I would look at the simplified core (i.e. `-ddump-simpl`) to verify that this is the case.

include/stacktracetools.h
leftaroundabout: I don't suppose it would be ok for `vector` to actually depend on `call-stack`?

feuerbach:
FWIW, I rebased @leftaroundabout's patches on top of the current master here: https://github.com/feuerbach/vector/tree/feature/ghc-callstack

Let me know if there's anything else I can do to help this get merged.

cartazio:
Ooo. Nice!  I’ll take a look soon.  Everything is a tad slow what with the
plague, but this piques my interest.

feuerbach:
Thanks @cartazio.

I've also added some missing HasCallStack annotations on my branch.

Annotations are still missing in the Fusion/* modules (Data/Vector/Fusion/Stream/Monadic.hs, Data/Vector/Fusion/Bundle.hs, Data/Vector/Fusion/Bundle/Monadic.hs); I'm not sure whether they should be there or not.

Bodigrim:
@feuerbach The idea looks good to me, but I do not like `stacktracetools.h` dangling around. I would very much prefer to copy-paste the incantation in every file, it's not that long. @lehins @Shimuuar what do you think?

Shimuuar:
On other hand repeating same thing is asking for problems when someone will forget to edit one of files or make a mistake. Still it doesn't solve duplication completely. There's `CHECK` defined in both vector.h and stacktrace.h

lehins:
My opinion on this PR is that the direction of relying on HasCallStack is the right way to go, however adding more CPP is not the right approach to get to the goal.

I've just submitted a PR that reduces CPP usage in favor of HasCallStack and uses `call-stack` package for ghc-7.10 to get backwards compatibility: #397  For now I only addressed the functions that directly throw errors, next step will be propagating HasCallStack constraint to the rest of partial functions in vector. Any feedback is appreciated.

My sincere apologies for this PR to be dragged around for 4 years. I do think that this PR was an improvement, and should have gotten more attention.

Shimuuar:
Closing since #397 is merged

PR #185, opened 2017-10-04
Add Unbox instances for Identity and Const

fumieval:
They are trivial instances.


RyanGlScott:
Seems reasonable to me.

One small suggestion: since `Const` is poly-kinded on recent versions of `base`, you might want to consider adding:

```haskell
#if __GLASGOW_HASKELL__ >= 706
{-# LANGUAGE PolyKinds #-}
#endif
```

at the top so that the corresponding instances in `vector` will be poly-kinded as well.

RyanGlScott:
LGTM.

PR #187, opened 2017-10-19
Use SSE2 for floating-point operations on i386

### LINKED ISSUE #186: Tests fail on i386 architecture
This causes *vector* to fail to build on Debian. You can find the build logs [here](https://buildd.debian.org/status/fetch.php?pkg=haskell-vector&arch=i386&ver=0.12.0.1-1&stamp=1505396428&raw=0). Could you please have a look at this?

iliastsi:
Closes #186: Tests fail on i386 architecture

### DIFF
--- a/vector.cabal
+++ b/vector.cabal
@@ -153,6 +153,9 @@ Library
 
   Ghc-Options: -O2 -Wall
 
+  if arch(i386)
+    Ghc-Options: -msse2
+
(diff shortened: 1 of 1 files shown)


cartazio:
ghc 8.10 onwards only supports sse2 mode, so this isn't needed for current ghc :)

PR #190, opened 2017-12-17
Add argument haddock for `move`

markus1189:
The docs are missing the `source` and `target` haddock in the `move` function.  Added them because the order is rather counter intuitive for me

### 24 Pull Requests (https://24pullrequests.com) ###
https://24pullrequests.com


RyanGlScott:
Thanks! (And sorry for taking so long to get around to looking at this!)

PR #191, opened 2017-12-18
Add `dropRight` and `takeRight`

markus1189:
I missed those functions recently and added them.

This was quite a bit of boilerplate 😞

### DIFF
--- a/Data/Vector/Generic.hs
+++ b/Data/Vector/Generic.hs
@@ -30,8 +30,9 @@ module Data.Vector.Generic (
   unsafeIndexM, unsafeHeadM, unsafeLastM,
 
   -- ** Extracting subvectors (slicing)
-  slice, init, tail, take, drop, splitAt,
+  slice, init, tail, take, takeEnd, drop, dropEnd, splitAt,
   unsafeSlice, unsafeInit, unsafeTail, unsafeTake, unsafeDrop,
+  unsafeTakeEnd, unsafeDropEnd,
 
   -- * Construction
 
@@ -415,6 +416,16 @@ take :: Vector v a => Int -> v a -> v a
 take n v = unsafeSlice 0 (delay_inline min n' (length v)) v
   where n' = max n 0
 
+-- | /O(1)/ Yield the last @n@ elements without copying. The vector may
+-- contain less than @n@ elements in which case it is returned unchanged.
+takeEnd :: Vector v a => Int -> v a -> v a
+{-# INLINE_FUSED takeEnd #-}
+takeEnd n v = unsafeSlice i (delay_inline min n' m) v
+  where
+    n' = max n 0
+    m = length v
+    i = delay_inline max 0 (m - n')
+
 -- | /O(1)/ Yield all but the first @n@ elements without copying. The vector may
 -- contain less than @n@ elements in which case an empty vector is returned.
 drop :: Vector v a => Int -> v a -> v a
@@ -423,6 +434,15 @@ drop n v = unsafeSlice (delay_inline min n' len)
                        (delay_inline max 0 (len - n')) v
   where n' = max n 0
         len = length v
+
+-- | /O(1)/ Yield all but the last @n@ elements without copying. The vector may
+-- contain less than @n@ elements in which case an empty vector is returned.
+dropEnd :: Vector v a => Int -> v a -> v a
+{-# INLINE_FUSED dropEnd #-}
+dropEnd n v = unsafeSlice 0 (delay_inline max 0 (m - n')) v
+  where
+    n' = max n 0
+    m = length v
 
 -- | /O(1)/ Yield the first @n@ elements paired with the remainder without copying.
 --
@@ -466,11 +486,24 @@ unsafeTake :: Vector v a => Int -> v a -> v a
 {-# INLINE unsafeTake #-}
 unsafeTake n v = unsafeSlice 0 n v
 
+-- | /O(1)/ Yield the last @n@ elements without copying. The vector must
+-- contain at least @n@ elements but this is not checked.
+unsafeTakeEnd :: Vector v a => Int -> v a -> v a
+{-# INLINE unsafeTakeEnd #-}
+unsafeTakeEnd n v = unsafeSlice (m - n) n v
+  where m = length v
+
 -- | /O(1)/ Yield all but the first @n@ elements without copying. The vector
 -- must contain at least @n@ elements but this is not checked.
 unsafeDrop :: Vector v a => Int -> v a -> v a
 {-# INLINE unsafeDrop #-}
 unsafeDrop n v = unsafeSlice n (length v - n) v
+
+-- | /O(1)/ Yield all but the last @n@ elements without copying. The vector
+-- must contain at least @n@ elements but this is not checked.
+unsafeDropEnd :: Vector v a => Int -> v a -> v a
+{-# INLINE unsafeDropEnd #-}
+unsafeDropEnd n v = unsafeSlice 0 (length v - n) v
 
 {-# RULES
 
--- a/Data/Vector/Generic/Mutable.hs
+++ b/Data/Vector/Generic/Mutable.hs
@@ -21,8 +21,9 @@ module Data.Vector.Generic.Mutable (
   length, null,
 
   -- ** Extracting subvectors
-  slice, init, tail, take, drop, splitAt,
+  slice, init, tail, take, takeEnd, drop, dropEnd, splitAt,
   unsafeSlice, unsafeInit, unsafeTail, unsafeTake, unsafeDrop,
+  unsafeTakeEnd, unsafeDropEnd,
 
   -- ** Overlapping
   overlaps,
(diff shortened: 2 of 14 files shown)


OlivierSohn:
I suggest a change in names : `dropEnd`, `takeEnd`, I find it is less ambiguous than `dropRight`, `takeRight` (this is also the naming used in `text`package : http://hackage.haskell.org/package/text-1.2.2.1/docs/Data-Text.html#v:dropEnd)

markus1189:
Sure, done!

OlivierSohn:
Should the unsafe versions be added, too? I see that for example `take` and `drop` have unsafe versions.

Data/Vector/Generic.hs
```diff
@@ -415,6 +415,16 @@ take :: Vector v a => Int -> v a -> v a
...
 
+-- | /O(1)/ Yield the last @n@ elements without copying. The vector may
+-- contain less than @n@ elements in which case it is returned unchanged.
+takeEnd :: Vector v a => Int -> v a -> v a
+{-# INLINE_FUSED takeEnd #-}
+takeEnd n v = unsafeSlice i (min n' m) v
+  where
+    n' = max n 0
+    m = length v
+    i = max 0 (m - n)
```
basile-henry: What happens if `n` is negative? What about doing `max 0 (m - n')` instead? :smile:

Data/Vector/Generic.hs
```diff
@@ -415,6 +415,16 @@ take :: Vector v a => Int -> v a -> v a
 take n v = unsafeSlice 0 (delay_inline min n' (length v)) v
   where n' = max n 0
 
+-- | /O(1)/ Yield the last @n@ elements without copying. The vector may
+-- contain less than @n@ elements in which case it is returned unchanged.
+takeEnd :: Vector v a => Int -> v a -> v a
+{-# INLINE_FUSED takeEnd #-}
+takeEnd n v = unsafeSlice i (min n' m) v
```
OlivierSohn: Based on the use of `delay_inline ` in `take`function, I think this should be:

`unsafeSlice i (delay_inline min n' m) v`

Data/Vector/Generic.hs
OlivierSohn: Based on the use of `delay_inline` in `take` function, I think this should be:

`i = delay_inline max 0 (m-n')`

Data/Vector/Generic.hs
```diff
@@ -424,6 +434,15 @@ drop n v = unsafeSlice (delay_inline min n' len)
   where n' = max n 0
         len = length v
 
+-- | /O(1)/ Yield all but the last @n@ elements without copying. The vector may
+-- contain less than @n@ elements in which case an empty vector is returned.
+dropEnd :: Vector v a => Int -> v a -> v a
+{-# INLINE_FUSED dropEnd #-}
+dropEnd n v = unsafeSlice 0 (max 0 (m - n')) v
```
OlivierSohn: Here also, I think we need to add a `delay_inline` for the `max`

Data/Vector/Generic/Mutable.hs
```diff
@@ -518,13 +518,28 @@ take :: MVector v a => Int -> v s a -> v s a
 {-# INLINE take #-}
 take n v = unsafeSlice 0 (min (max n 0) (length v)) v
 
+takeEnd :: MVector v a => Int -> v s a -> v s a
+{-# INLINE takeEnd #-}
+takeEnd n v = unsafeSlice i (min n' m) v
+  where
+    n' = max n 0
+    m = length v
+    i = max 0 (m - n)
```
OlivierSohn: I guess it's safer to write `max 0 (m-n')`

old-testsuite/Testsuite/Vector/AsList.hs
```diff
@@ -18,6 +18,12 @@ instance Modelled (Vector a) where
   model = V.toList
   unmodel = V.fromList
 
+naiveTakeEnd :: Int -> [a] -> [a]
```
OlivierSohn: This test function is duplicated to be used in two files, I would rather put it in a separate module to avoid duplication, like the existing `Utilities`module.

old-testsuite/Testsuite/Vector/AsList.hs
```diff
@@ -18,6 +18,12 @@ instance Modelled (Vector a) where
...
+naiveTakeEnd n = reverse . take n . reverse
+
+naiveDropEnd :: Int -> [a] -> [a]
```
OlivierSohn: This one, too.

Data/Vector/Generic.hs
markus1189: Done

Data/Vector/Generic.hs
markus1189: Done

Data/Vector/Generic.hs
markus1189: Done

Data/Vector/Generic.hs
markus1189: Done

Data/Vector/Generic/Mutable.hs
markus1189: Agreed. Done

old-testsuite/Testsuite/Vector/AsList.hs
markus1189: Moved to Utilities

old-testsuite/Testsuite/Vector/AsList.hs
markus1189: Moved to Utilities

markus1189:
Thanks for the thorough review!  Next step is to add the unsafe versions as well then.  I wish it would be less boilerplate-y though

OlivierSohn:
@markus1189 Yes, it looks like it will be a lot of copy / paste / change one word stuff :)

Maybe you can ask a maintainer of the library if it's necessary or not, before doing the work of adding the unsafe versions. (I'm not a maintainer myself)

markus1189:
I added the `unsafe` versions.  I hope I did not make any copy & paste errors ;)

markus1189:
I am willing to continue to work on this, but before I do I would like to have some feedback from the maintainers whether this is wanted at all. 😉

cartazio:
Probably?
I / Dan / Ryan are eyeballs deep in work / oss backlog (or at least I am :))

Hit me up this Saturday morning/ midday on freenode on one of the Haskell
channels (#numerical-haskell may have some folks who can also help review )
if I’ve not looked at it by then. Or Twitter. Handles there are
respectively Carter and cartazio.

Cheers.

markus1189:
Okay I'll try to fix the issues until Saturday.  I don't know yet how much
time I'll have on Saturday, but for the records: what's your timezone? :D

I'm in CET (Frankfurt, Germany)

cartazio:
NYC time :)

markus1189:
I kind of gave up on writing the prop for the naive version.  It seems like there is quite some customizing going on in the QuickCheck tests.

Some help would be great there ;)

@OlivierSohn Fixed the issues you raised, thanks a lot for the in-depth reviews!

cartazio:
possibly dumb question, and this might just be me missing a difference ... but how is takeEnd different from drop?, just a difference in in counting from front of vector vs end of vector?

this seems to be a useful difference, and the key one that motivates the changes?

cartazio:
so one possible question about this for code simplification: is there any reason to not define takeEnd in terms of drop, and dropEnd in terms of take? (one reason not to would be if the generated core was less awesome under O2 or something)

cartazio:
also you duplicated an Orphan, https://travis-ci.org/haskell/vector/jobs/330841124#L908-L918 
use the orphan module for that

tests/Tests/Vector.hs
```diff
@@ -71,6 +71,14 @@ type VectorContext  a v = ( Eq (v a), Show (v a), Arbitrary (v a), CoArbitrary (
 
 -- TODO: test non-IVector stuff?
 
+#if !MIN_VERSION_base(4,7,0)
+instance Foldable ((,) a) where
```
cartazio: ORPHAN, see Data.Orphan in tests

markus1189:
@cartazio Yes indeed, `takeEnd` is `drop` but avoids the calculation of correct offsets (equivalently for `take` and `dropEnd`)

I missed those handy functions recently and thought to add them (this was in 24pullrequests btw :) )

This got a little out of hand, I thought it would be much less stuff to do, though 😉 

Will check the orphans

cartazio:
I guess my point is that if those overheads/extra calculations are optimized away under O2 level optimization and have equivalent fusion characteristics, may as well do the simpler definition in terms of preexisting high level operations. I would be surprised if the optimized core isn't essentially the same. (worth a bug report even :) )

OlivierSohn:
> also you duplicated an Orphan, https://travis-ci.org/haskell/vector/jobs/330841124#L908-L918
use the orphan module for that

I see the same error on my pull request, strange... is it possible to launch a build of master to see if the build is still ok?

cartazio:
https://travis-ci.org/haskell/vector/builds/331565078

cartazio:
master is fine, this patch is the issue :)

cartazio:
i restarted the PR build so we can look again

OlivierSohn:
can you restart the one on #196 please ? I have the exact same error

markus1189:
Sorry but I have to abandon this as I don't come around to do it.  If anyone wants to pick it up, feel free ;)

noughtmare:
I think it would be better to use the names `dropL`, `dropR`, `takeL`, and `takeR` (like what the Data.Sequence module in the containers library does). Because vectors have no bias to left or right. Text (and arguably bytestrings) do have such a bias, so the asymmetric naming scheme makes sense for those types.

lehins:
@noughtmare Would you like to create an issue with a link to this PR, so it doesn't get forgotten and someone with time on their hands can pick it up.

PR #193, opened 2017-12-29
fix typo in construct[r]N examples

int-e:
(there are extra `f` call in the results)


RyanGlScott:
Thanks for doing this.

It looks like there are more occurrences of `constructN` and `constructrN` (with the corresponding Haddock typos) in `Data.Vector.Generic`, `Data.Vector.Primitive`, and `Data.Vector.Storable`. Can you apply the fix in those places as well?

int-e:
Wow, I didn't expect there to be so many copies of this interface. Anyway, done.

PR #205, opened 2018-04-03
Add new Data instances for immutable vectors

andy-morris:
I included a test for `gfoldl`, but not for `gunfold` because I don't really know how to use it. If someone has any pointers, I'll be happy to add one for it too.

### DIFF
--- a/Data/Vector/Generic.hs
+++ b/Data/Vector/Generic.hs
@@ -165,7 +165,7 @@ module Data.Vector.Generic (
   liftShowsPrec, liftReadsPrec,
 
   -- ** @Data@ and @Typeable@
-  gfoldl, gunfold, dataCast, mkTypeConstr
+  gfoldl, gunfold, dataCast, mkVecType, mkVecConstr
 ) where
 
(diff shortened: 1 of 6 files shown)


Data/Vector/Generic.hs
```diff
@@ -2205,9 +2199,19 @@ gfoldl :: (Vector v a, Data a)
 {-# INLINE gfoldl #-}
 gfoldl f z v = z fromList `f` toList v
 
-mkType :: String -> DataType
-{-# INLINE mkType #-}
-mkType = mkNoRepType
+mkTypeConstr :: String -> (DataType, Constr)
```
RyanGlScott: Tupling these together feels a mite strange. Why not separate these into two separate functions which return `DataType` and `Constr` respectively? Then you wouldn't have to manually project out of it with `fst` and `snd` everywhere.

andy-morris: I only tupled them because they're mutually recursive, but I know that's a bit of a silly reason. I'll separate them this evening when I get a chance.

PR #209, opened 2018-04-23
Add function for converting Array to Vector

andrewthad:


cartazio:
@andrewthad can you add a unit test?

lehins:
@andrewthad It looked like this PR stalled a bit so I took it upon myself to add the unit test and extend the PR a bit. Would you mind giving it a review, I'd really appreciate it?

@Bodigrim and @Shimuuar This PR adds functions for conversion to boxed arrays and back, if you have time, could you take a look at it. There is nothing complicated, but one notable change is the bump in a dependency on minimum version of primitive to 0.6.4.0, because somehow that is the first version that provides `sizeofArray` and `sizeofMutableArray`

Data/Vector.hs
```diff
@@ -1776,8 +1776,19 @@ fromListN = G.fromListN
...
+{-# INLINE fromArray #-}
 fromArray x = Vector 0 (sizeofArray x) x
 
+-- | /O(n)/ Convert a vector to an array.
+toArray :: Vector a -> Array a
+{-# INLINE toArray #-}
+toArray vec@(Vector offset size arr)
+  | offset == 0 && size == sizeofArray arr = arr
+  | otherwise = runST $ do
+      mvec <- unsafeThaw vec
```
andrewthad, line 1788: `unsafeThaw` is not needed. Using `Data.Primitive.Array.cloneArray` (backed by the `cloneArray#` primop) performs a little better because it only sets the closure type once (and on GHC 8.10 and up, avoids a write barrier when the concurrent collector is on).

Data/Vector/Mutable.hs
```diff
@@ -428,3 +431,23 @@ unsafeMove = G.unsafeMove
...
+fromMutableArray marr = do
+  let !size = sizeofMutableArray marr
+  mvec <- G.basicUnsafeNew size
+  G.basicUnsafeCopy mvec (MVector 0 size marr)
+  return mvec
+
+-- | /O(n)/ Make a copy of a mutable vector into a new mutable array.
+toMutableArray :: PrimMonad m => MVector (PrimState m) a -> m (MutableArray (PrimState m) a)
+{-# INLINE toMutableArray #-}
+toMutableArray (MVector offset size marrSrc) = do
```
andrewthad, line 450: Using `Data.Primitive.Array.cloneMutableArray` avoid the needless `memset` that `newArray` uses here.

Data/Vector/Mutable.hs
```diff
@@ -428,3 +431,23 @@ unsafeMove = G.unsafeMove
...
 {-# INLINE nextPermutation #-}
 nextPermutation = G.nextPermutation
+
+-- Conversions - Arrays
+-- -----------------------------
+
+-- | /O(n)/ Make a copy of a mutable array to a new mutable vector.
+fromMutableArray :: PrimMonad m => MutableArray (PrimState m) a -> m (MVector (PrimState m) a)
+{-# INLINE fromMutableArray #-}
+fromMutableArray marr = do
```
andrewthad, line 441: I don't really know how I feel about this function. Whenever I've wanted `fromMutableArray` before, I've wanted it to alias the original array, not to make a copy. I think I would prefer something like:

    fromMutableArray :: MutableArray (PrimState m) a -> Int -> Int -> MVector (PrimState m) a

Which explicitly accepts the offsets (and is just a wrapper for the data constructor). Such an interface is more powerful, since it would allow me to temporarily or permanently "step up" into an `MVector` during a computation if there is some machinery there that I happen to need. It's possible to provide both interfaces though.

Data/Vector.hs
```diff
@@ -1776,8 +1776,19 @@ fromListN = G.fromListN
 
 -- | /O(1)/ Convert an array to a vector.
 fromArray :: Array a -> Vector a
+{-# INLINE fromArray #-}
```
andrewthad, line 1779: Although I originally wrote this function, I now feel that accepting offset and length would be better here. This gives the user a little more power. Of course, the drawback is that the user can segfault the program if they give bad boundaries.

Data/Vector.hs
lehins: Good point, I wanted to use `cloneArray`, but saw that it wasn't available in `primitive-0.5.0.1` (current min version), but it is definitely better to just bump up the version in order to get the proper implementation.

Data/Vector/Mutable.hs
lehins: Same thing here, good catch.

Data/Vector/Mutable.hs
lehins: I disagree on this one. `MVector` is exported, so the function you are talking about would be just a synonym for the constructor. If a user wants to manually deal with dangerous indices and offsets then we won't prevent him, but we should not promote it by creating a separate function for such functionality.

Data/Vector.hs
lehins: Same thing here, constructor is available for such use cases. It is better to follow it with a slice function, i.e.:
* safe: `slice offset size . fromArray`
* unsafe: `unsafeSlice offset size . fromArray`

Shimuuar: I agree with @lehins here. Functions that could violate memory safety should be clearly marked as such.

lehins:
@andrewthad I updated the implementation to your suggestions to use clone functions. Can you do a sanity check?

Bodigrim:
I'm not quite sure how useful this is. Does anyone have a particular use case in mind?

lehins:
@Bodigrim I do. `Array` and `MutableArray` serve as wrappers around GHC primitive values as such they can serve as very useful intermediaries for conversion between different libraries that don't depend on vector but would like to provide functionality to convert to vector and back. `primitive`, `contiguous`, `massiv, `basement` etc.

People who rely on vector so much tend to forget that there are other libraries out there ;)

Without these functions users have to either convert manually and make sure to get the offset and size correctly, or use a more inefficient approach like `generate`+`index`

PR #212, opened 2018-05-07
uncons, unsnoc

strake:
I often find these useful.


strake:
ping

Data/Vector/Generic.hs
```diff
@@ -432,6 +434,14 @@ splitAt n v = ( unsafeSlice 0 m v
       n'  = max n 0
       len = length v
 
+-- | /O(1)/ Yield the 'head' and 'tail' of the vector, or 'Nothing' if empty.
+uncons :: Vector v a => v a -> Maybe (a, v a)
+uncons xs = flip (,) (tail xs) <$> xs !? 0
```
RyanGlScott: Wouldn't it be more efficient to use `unsafeTail` here, since you know in this case that the `Vector` is non-empty?

Data/Vector/Generic.hs
```diff
@@ -432,6 +434,14 @@ splitAt n v = ( unsafeSlice 0 m v
...
+
+-- | /O(1)/ Yield the 'last' and 'init' of the vector, or 'Nothing' if empty.
+unsnoc :: Vector v a => v a -> Maybe (v a, a)
+unsnoc xs = (,) (init xs) <$> xs !? (length xs - 1)
```
RyanGlScott: Similarly, `unsafeInit` here?

strake:
Done, PTAL

RyanGlScott:
LGTM.

@cartazio, thoughts?

Data/Vector/Generic.hs
```diff
@@ -432,6 +434,14 @@ splitAt n v = ( unsafeSlice 0 m v
       n'  = max n 0
       len = length v
 
+-- | /O(1)/ Yield the 'head' and 'tail' of the vector, or 'Nothing' if empty.
```
cartazio: why aren't we thinking about how to adapt the fusion codes for head/tail as found in Stream.Monadic and/or bundle?

https://github.com/haskell/vector/blob/master/Data/Vector/Fusion/Stream/Monadic.hs#L341
for one of them

cartazio:
1) i think this is a VALID / USEFUL contribution

2)  this seems like a solid natureal implementation if we ignore fusion.... but should we? (Ryan points out that eg the partition function does too, but should it?)

3) how about we figure this out by seeing if we can construct the bundle/stream monad combinator versions of this and compare them

4) (@RyanGlScott how many things in Vector.Generic dont have an impl in bundle or stream? we may need to review that ... perhaps on another ticket
)

cartazio:
roughly: ignoring fusion this looks good ... but if we ignore fusion we should just use Array right? :)

strake:
@cartazio I think `uncons` would make sense on `Stream`s, but am not sure how to write `unsnoc` efficiently; is the latter worth doing?

cartazio:
i think you can look at [stream-take](https://github.com/haskell/vector/blob/master/Data/Vector/Fusion/Stream/Monadic.hs#L361-L372 ) and [stream-last](https://github.com/haskell/vector/blob/master/Data/Vector/Fusion/Stream/Monadic.hs#L256-L274) definitions as starting points perhaps? 

I swear I saw some machinery for left vs right streams at some point in vector, but i'm not finding it at the moment

cartazio:
https://github.com/haskell/vector/blob/master/Data/Vector/Fusion/Stream/Monadic.hs#L143 has length, not srue if thats quite th right form though

cartazio:
i might be missing something obvious mind you :)

strake:
@cartazio any word on what needs to be done to accept this PR? (e.g. bundle ops)

cartazio:
Reminding me is a great way!  I was a bit snowed in intellectual for a
while so I’m climbing over large snow drifts metaphorically


The best way to pull me into prioritizing stuff is to chat with me on
freenode irc.

strake:
@cartazio ping

cartazio:
Derp!
Thx.  I’ll poke at this soon.

cartazio:
now that the most recent release has been done, i'm catching up on polishing backlog.

i'll look

lehins:
@strake I do support @cartazio push for thinking about fusion. 

With respect to `unsnoc` it is not feasible to implement it without either duplicated work or materializing the whole vector, because you'd require to iterate the whole stream in order to get to the last element. Therefore your `unsnoc` implementation is perfectly fine. 

As far as `uncons` in concerned we can convert a vector to a stream, grab the first element from that stream and construct another stream where we drop that first element. I could be wrong of course, since I haven't tried how fusion with rewrite rules will interact with such approach, but I think it should be fine.

```haskell
uncons :: Monad m => Stream m a -> m (Maybe (a, Stream m a))
uncons str@(Stream step t) = unconsMaybeLoop SPEC t
  where
    unconsMaybeLoop !_ s = do
      r <- step s
      case r of
        Yield x _ -> return $ Just (x, drop 1 str)
        Skip s'   -> unconsMaybeLoop SPEC s'
        Done      -> return Nothing
    {-# INLINE [0] unconsMaybeLoop #-}
{-# INLINE uncons #-}
```

An even more general and useful approach would be to implement `headMaybe` in the exact same manner and reuse that to implement `uncons`:

```haskell
headMaybe (Stream step t) = headMaybeLoop SPEC t
  where
    headMaybeLoop !_ s = do
      r <- step s
      case r of
        Yield x _ -> return $ Just x
        Skip s'   -> headMaybeLoop SPEC s'
        Done      -> return Nothing
    {-# INLINE [0] headMaybeLoop #-}
{-# INLINE headMaybe #-}
```

@strake  If you'd like to give it a shot that be great if not I'll try it out in a few days myself.

lehins:
Trying to integrate with fusion only makes it worse. Because we have constant time slicing consecutive uncons/unsonc are very efficient with current implementation.

lehins:
@strake Thank you for the PR. Sorry it took so long to get it merged.

PR #215, opened 2018-06-07
Make boxed Vector parameter representational

peddie:
This commit allows `Data.Vector.Vector`s to be `coerce`d.

### DIFF
--- a/Data/Vector.hs
+++ b/Data/Vector.hs
@@ -6,6 +6,9 @@
            , Rank2Types
            , BangPatterns
   #-}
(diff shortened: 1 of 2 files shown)


RyanGlScott:
I'm not sure I understand the purpose of these role annotations, as `Vector`'s role argument is _already_ representational. (And `MVector`'s arguments are already nominal and representational, respectively.)

peddie:
I'm sorry, I made a mistake.  Carter suggested that these roles might be at fault, but you're right; my problem with coercion must lie elsewhere.  Thanks for the quick reply.

Issue #223, opened 2018-10-27
Vector should be representational


idontgetoutmuch:
At the moment we get this

```
*Numeric.Sundials.CVode.ODE> (coerce ((V.fromList [1,2]) :: Vector CInt)) :: Vector Int
[8589934593,4426664329]
```

We should really get a type error.

idontgetoutmuch:
And by making Vector representational we get

```
*Data.Vector> (coerce ((Data.Vector.fromList [1,2]) :: Vector CInt)) :: Vector Int

<interactive>:9:2: error:
    • Couldn't match representation of type ‘GHC.Int.Int32’
                               with that of ‘Int’
        arising from a use of ‘coerce’
    • In the expression:
          (coerce ((Data.Vector.fromList [1, 2]) :: Vector CInt)) ::
            Vector Int
      In an equation for ‘it’:
          it
            = (coerce ((Data.Vector.fromList [1, 2]) :: Vector CInt)) ::
                Vector Int
```

RyanGlScott:
Fixed in #224.

PR #224, opened 2018-10-27
Fix https://github.com/haskell/vector/issues/223

idontgetoutmuch:


idontgetoutmuch:
Hopefully all is now well

RyanGlScott:
LGTM. Thanks!

ekmett:
Isn't representational too weak for safety? In general when you have a container that behaves based on the instances associated with a type (e.g. Map w/ Ord, etc) we want a nominal role for the argument. As it stands a user could still coerce one of them that was made representational. Having the same representation does not mean the Storable instances are compatible. They could well have different alignments or formats.

RyanGlScott:
Yes, that is a good observation. I'm somewhat reluctant to just slam on the brakes and make the arguments to `Primitive.Vector` and `Storable.Vector` nominal since it's possible that there is code in the wild that relies on `coerce`-ing underneath them (I would wager that @idontgetoutmuch probably has some examples of such code, in fact). Still, the safety argument is quite concerning...

Do you have an example of some code that would go wrong at runtime due to these roles being representational? That would make for a pretty compelling counterexample.

Shimuuar:
Totally contrived example for `Storable`. Data type with phantom type which mirrors C structure and phantom type encode layout (packed, standard C alignment,  etc.).

cartazio:
Storable needs to be nominal.  Because of binary encoding things.  Like
network order vs host order due to storable being for c interop

The semantics for prim is “the only instances should directly map to their
unlifted sibling”. Anything else should just be a new type.  Funkier / more
interesting things should be in unboxed.

Prim can be representational afaict.

Also this could have used more review before merge for that very reason...

Somehow didn’t see the thread this weekend :(

RyanGlScott:
My apologies for pushing it through quickly—I legitimately thought this was a no-brainer at the time. And in any case, this is a move in the right direction, since the roles used to be even more permissive than they are now! The remaining question is whether we should go further in restricting the roles.

@idontgetoutmuch, out of curiosity, which flavor of `Vector` were you `coerce`-ing under? (I'm not sure if it's the `Prim`- or `Storable`-based one.)

ekmett:
We could always offer a function ourself for the coercion, an `unsafeCoerce` if you will. ;)

RyanGlScott:
I see. Unfortunately, that's exactly the flavor of `Vector` that ought to be nominally roled, as I've now come to realize, which means that this code would stop working if we were to give it the role signature it really deserves :(

I think that if we are going to do this, we should follow @ekmett's advice and offer something like:

```haskell
unsafeVectorCoerce :: Coercible a b => Storable.Vector a -> Storable.Vector b
unsafeVectorCoerce = unsafeCoerce
```

As the name suggests, it would still be unsound to use this in general, but if you _do_ know what you're doing, then you can at least have some assurance that the type parameters are representationally equal.

cartazio:
Yeah. For storable anything ever, you can’t convert without actually
looking at the exact actual binary / memory encoding

idontgetoutmuch:
@RyanGlScott  Yes I feared as much when I looked at my code but I was beguiled by the the word "safe" in the documentation. Yes I should use `unsafeVectorCoerce` and presumably have a test such that should ghc ever change the representation for `Double` it will get caught.

RyanGlScott:
I've submitted #235 to make `Storable` vectors nominally roled.

PR #226, opened 2018-11-05
Add mapMaybeM and imapMaybeM

### LINKED ISSUE #183: mapMaybeM
Can I add `mapMaybeM` to `vector`? The type signature would be:

    mapMaybeM :: (a -> m (Maybe b)) -> Vector a -> m (Vector b)

Although the list-equivalent of `mapMaybeM` is not in `base`, it's a pretty common thing for people to write. I can see from some googling that it exists in [`monad-extras`](https://hackage.haskell.org/package/monad-extras-0.6.0/docs/Control-Monad-Extra.html#v:mapMaybeM), [`extra`](https://hackage.haskell.org/package/extra-1.6/docs/Control-Monad-Extra.html#v:mapMaybeM), and in [`stack`](https://hackage.haskell.org/package/stack-1.0.0/docs/src/Data.Maybe.Extra.html). I bring this up because I think it highlights the utility of such a function and the common expected semantics people seem to have for it.

treeowl:
Add

```haskell
mapMaybeM :: Monad m => (a -> m (Maybe b)) -> Vector a -> m (Vector b)
imapMaybeM :: Monad m => (Int -> a -> m (Maybe b)) -> Vector a -> m (Vector b)
```

`mapMaybeM` is similar to `wither`, but the stream fusion framework
seems to require that we use a `Monad` constraint rather than an
`Applicative` one to get good performance. `imapMaybeM` is the
indexed variant.

Resolves #183

### DIFF
--- a/Data/Vector/Generic.hs
+++ b/Data/Vector/Generic.hs
@@ -98,7 +98,7 @@ module Data.Vector.Generic (
   -- ** Filtering
   filter, ifilter, uniq,
   mapMaybe, imapMaybe,
-  filterM,
+  filterM, mapMaybeM, imapMaybeM,
   takeWhile, dropWhile,
 
   -- ** Partitioning
@@ -1333,6 +1333,14 @@ imapMaybe f = unstream
 filterM :: (Monad m, Vector v a) => (a -> m Bool) -> v a -> m (v a)
 {-# INLINE filterM #-}
 filterM f = unstreamM . Bundle.filterM f . stream
+
+mapMaybeM :: (Monad m, Vector v a, Vector v b) => (a -> m (Maybe b)) -> v a -> m (v b)
+{-# INLINE mapMaybeM #-}
+mapMaybeM f = unstreamM . Bundle.mapMaybeM f . stream
+
+imapMaybeM :: (Monad m, Vector v a, Vector v b)
+      => (Int -> a -> m (Maybe b)) -> v a -> m (v b)
+imapMaybeM f = unstreamM . Bundle.mapMaybeM (\(i, a) -> f i a) . Bundle.indexed . stream
 
 -- | /O(n)/ Yield the longest prefix of elements satisfying the predicate
 -- without copying.
--- a/Data/Vector.hs
+++ b/Data/Vector.hs
@@ -114,8 +114,8 @@ module Data.Vector (
 
   -- ** Filtering
   filter, ifilter, uniq,
-  mapMaybe, imapMaybe,
-  filterM,
+  mapMaybe, imapMaybe, imapMaybeM,
+  filterM, mapMaybeM,
   takeWhile, dropWhile,
 
   -- ** Partitioning
@@ -1213,6 +1213,14 @@ imapMaybe = G.imapMaybe
 filterM :: Monad m => (a -> m Bool) -> Vector a -> m (Vector a)
 {-# INLINE filterM #-}
 filterM = G.filterM
+
+mapMaybeM :: Monad m => (a -> m (Maybe b)) -> Vector a -> m (Vector b)
+{-# INLINE mapMaybeM #-}
+mapMaybeM = G.mapMaybeM
+
+imapMaybeM :: Monad m => (Int -> a -> m (Maybe b)) -> Vector a -> m (Vector b)
+{-# INLINE imapMaybeM #-}
+imapMaybeM = G.imapMaybeM
 
 -- | /O(n)/ Yield the longest prefix of elements satisfying the predicate
 -- without copying.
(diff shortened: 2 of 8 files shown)


Data/Vector/Generic.hs
```diff
@@ -1334,6 +1334,14 @@ filterM :: (Monad m, Vector v a) => (a -> m Bool) -> v a -> m (v a)
 {-# INLINE filterM #-}
 filterM f = unstreamM . Bundle.filterM f . stream
 
+mapMaybeM :: (Monad m, Vector v a, Vector v b) => (a -> m (Maybe b)) -> v a -> m (v b)
+{-# INLINE mapMaybeM #-}
+mapMaybeM f = unstreamM . Bundle.mapMaybeM f . stream
+
+imapMaybeM :: (Monad m, Vector v a, Vector v b)
```
RyanGlScott: Is it worthwhile to mark `imapMaybeM` as `INLINE`? (Genuine question—I'm not sure if it would be beneficial here or not.)

Data/Vector/Generic.hs
```diff
@@ -1334,6 +1334,14 @@ filterM :: (Monad m, Vector v a) => (a -> m Bool) -> v a -> m (v a)
...
+      => (Int -> a -> m (Maybe b)) -> v a -> m (v b)
+imapMaybeM f = unstreamM . Bundle.mapMaybeM (\(i, a) -> f i a) . Bundle.indexed . stream
```
RyanGlScott: Other indexed functions in this module seem to use `uncurry f` instead of `(\(i, a) -> f i a)`—perhaps it would be worth sticking to this convention? (I'm not sure if using one or the other has any actual ramifications on strictness properties.)

Data/Vector/Generic.hs
treeowl: I never use `uncurry` because it introduces laziness I virtually never need. I bet the compiler often fixes that, but I see no reason to tempt fate.

Data/Vector/Generic.hs
treeowl: I never use `uncurry` because it introduces laziness I virtually never need. I bet the compiler often fixes that, but I see no reason to tempt fate.

Data/Vector/Generic.hs
treeowl: I sure hope that helps. @bgamari has pointed out some trouble fusing `unstreamM`, but we surely want to fuse on the `stream` side.

cartazio:
Ooo.  This looks like it’s a good idea.  The wither style api pattern is a
great one.  And it does sound like something that should play nice with
stream fusion.

I’ll need to go over this code very thoroughly but I’m very much in favor
of this addition

treeowl:
@cartazio, it actually shouldn't require terribly much review. The whole thing is an extremely gentle modification of `filterM`.

treeowl:
FYI, the next release of `witherable` will offer a `witherM` method in `Witherable` to take advantage of this sort of thing.

fumieval:
Any updates? The change looks good

cartazio:
It’s top of my holiday vacation list :)

Data/Vector/Generic.hs
chessai: i wonder if the `uncurry` thing is actually necessary in general. i tend to prefer not using it as well.

cartazio:
this is an often requested function, if we can add it to part of the test suite that would make me game for inclusion

cartazio:
@Shimuuar @lehins any code review suggestions or ideas?

cartazio:
(inclusion for next release,)

lehins:
Adding documentation to those new functions would be a great idea. Otherwise LGTM

cartazio:
ok, i'm game for merging this in as long as we gate the release on a contrib of docs and tests from whomever

Shimuuar:
Only missing piece for this PR is documentation. I think easiest approach is to merge it as it is and I'll write haddock for functions later

lehins:
@Shimuuar if you'd rather not merge it into master without documentation, we could create a temporary branch from master for this PR and merge it into that branch, where you could finish up the doc. All that is necessary is creating a branch and changing the target of this PR

Shimuuar:
I propose to merge it _without_ documentation and it later separately. Seems like easiest approach

treeowl:
Do I need to write documentation today? Doesn't sound like a big deal.

Shimuuar:
No hurry. It's open for more than year. It could wait for few days

Shimuuar:
@treeowl Could you please add haddocks and changelog entry? So this PR could be finally merged

Shimuuar:
Superseded by #333 which is this PR rebased on top of master and added haddocks

PR #227, opened 2018-11-08
Document all instances of `slice` properly

hpdeifel:
Since the argument order of `slice` isn't apparent from the types,
it's especially important to document that.


RyanGlScott:
Good catch—these should have been kept in sync with the Haddocks for `Data.Vector.Generic.slice`. Thanks for fixing this.

cartazio:
Yay!  Good work

Issue #216, opened 2018-06-19
no instance of NFData1 Vector


ghorn:
I think it might look something like:
```haskell
instance NFData1 V.Vector where
  liftRnf arnf = V.foldl' (\_ x -> arnf x) ()
  {-# INLINEABLE liftRnf #-}
```
Maybe you want to define `NFData` in terms of `NFData1`

RyanGlScott:
This sounds reasonable to me. Care to submit a patch?

peddie:
See https://github.com/haskell/vector/pull/233 for a patch.

RyanGlScott:
Fixed in #233.

PR #233, opened 2018-12-28
Add NFData1 instance for boxed Vectors

peddie:
This addresses https://github.com/haskell/vector/issues/216

### DIFF
--- a/Data/Vector.hs
+++ b/Data/Vector.hs
@@ -169,7 +169,12 @@ import           Data.Vector.Mutable  ( MVector(..) )
 import           Data.Primitive.Array
 import qualified Data.Vector.Fusion.Bundle as Bundle
 
-import Control.DeepSeq ( NFData, rnf )
+import Control.DeepSeq ( NFData(rnf)
+#if MIN_VERSION_deepseq(1,4,3)
+                       , NFData1(liftRnf)
+#endif
+                       )
+
 import Control.Monad ( MonadPlus(..), liftM, ap )
 import Control.Monad.ST ( ST )
 import Control.Monad.Primitive
@@ -219,11 +224,18 @@ data Vector a = Vector {-# UNPACK #-} !Int
                        {-# UNPACK #-} !(Array a)
         deriving ( Typeable )
 
+liftRnfV :: (a -> ()) -> Vector a -> ()
+liftRnfV elemRnf = foldl' (\_ -> elemRnf) ()
+
 instance NFData a => NFData (Vector a) where
-    rnf (Vector i n arr) = rnfAll i
-        where
-          rnfAll ix | ix < n    = rnf (indexArray arr ix) `seq` rnfAll (ix+1)
-                    | otherwise = ()
+  rnf = liftRnfV rnf
+  {-# INLINEABLE rnf #-}
+
+#if MIN_VERSION_deepseq(1,4,3)
+instance NFData1 Vector where
+  liftRnf = liftRnfV
+  {-# INLINEABLE liftRnf #-}
+#endif
 
 instance Show a => Show (Vector a) where
   showsPrec = G.showsPrec
--- a/Data/Vector/Primitive.hs
+++ b/Data/Vector/Primitive.hs
@@ -147,7 +147,11 @@ import qualified Data.Vector.Fusion.Bundle as Bundle
 import           Data.Primitive.ByteArray
 import           Data.Primitive ( Prim, sizeOf )
 
-import Control.DeepSeq ( NFData(rnf) )
+import Control.DeepSeq ( NFData(rnf)
+#if MIN_VERSION_deepseq(1,4,3)
+                       , NFData1(liftRnf)
+#endif
+                       )
 
 import Control.Monad ( liftM )
 import Control.Monad.ST ( ST )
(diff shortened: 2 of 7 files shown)


peddie:
I also have a simpler patch adding the instance for immutable `Storable` vectors if there's interest.

Data/Vector.hs
```diff
@@ -169,7 +169,12 @@ import           Data.Vector.Mutable  ( MVector(..) )
 import           Data.Primitive.Array
 import qualified Data.Vector.Fusion.Bundle as Bundle
 
-import Control.DeepSeq ( NFData, rnf )
+import Control.DeepSeq ( NFData(rnf)
+#if __GLASGOW_HASKELL__ >= 821
```
RyanGlScott: Use `#if MIN_VERSION_deepseq(1,4,3)` here instead of `#if __GLASGOW_HASKELL__ >= 821`, since it's entirely possible to use an `NFData1`-equipped `deepseq` with older versions of GHC than 8.2.

Data/Vector.hs
```diff
@@ -219,11 +224,24 @@ data Vector a = Vector {-# UNPACK #-} !Int
...
+  rnf = liftRnf rnf
+  {-# INLINEABLE rnf #-}
+
+instance NFData1 Vector where
+  liftRnf elemRnf = foldl' (\_ -> elemRnf) ()
+  {-# INLINEABLE liftRnf #-}
+#endif
+
+#if __GLASGOW_HASKELL < 821
+instance NFData a => NFData (Vector a) where
```
RyanGlScott: Having two entirely separate code paths here is a little unfortunate. I would prefer to abstract out the common bits into a top-level function (which isn't exported), which we might name:

```haskell
liftRnfVector :: (a -> ()) -> Vector a -> ()
liftRnfVector = ...
```

Once this is defined, we can implement the instances like so:

```haskell
instance NFData a => NFData (Vector a) where
  rnf = liftRnf rnf

#if MIN_VERSION_deepseq(1,4,3)
instance NFData1 Vector where
  liftRnf = liftRnfVector
#endif
```

This ensures that we use the same code paths everywhere, and it keeps CPP to a minimum.

Data/Vector.hs
peddie: You got it!

Data/Vector.hs
peddie: No problem.

peddie:
@RyanGlScott thanks for your quick review during the holidays.  I've added all the instances I can, along with the changes from your helpful suggestions.  

The `NFData` instance for mutable boxed vector is commented out as "seems unsafe," but [the Trac link given in the comments](http://trac.haskell.org/vector/ticket/54) is dead.  Do you know anything about this?

cartazio:
I’ll have a look in a few day, just got back from a week long fam trip.
I’ll be doing a bunch of vector review next week one way or another :)

peddie:
Please let me know if there's anything else you'd like me to do to get this patch ready to merge.

cartazio:
this looks good. I think i spoke with Peddie last weekend on IRC to walk through likely meanings of unsafety wrt mutable boxed vectors : namely thunk evaluation in a multi threaded setting

RyanGlScott:
If you're happy with this, then so am I. I'll go ahead and merge this.

> I think i spoke with Peddie last weekend on IRC to walk through likely meanings of unsafety wrt mutable boxed vectors : namely thunk evaluation in a multi threaded setting

If there are issues with the (already existing) `NFData` instances for mutable boxed vectors, then perhaps that should be dealt with in a separate commit. (Although I don't quite understand how this unsafe behavior can be triggered using only `rnf`.)

peddie:
@RyanGlScott I don't think the `Storable` and `Primitive` instances suffer from the same issues, as they are just flat memory (nothing is boxed).

cartazio:
Indeed as @peddie and I hashed out, boxed mutable vectors are the only family of vectors in the `vector` package which aren't some flavore of unboxed flat memory, and thus the only storage class that has the possiblity of holding thunks. RNF on every other class of mutable vector is just a mechanism for doing WHNF on the data constructor of the type. 

(i'm not sure if we're correctly strict enough on the unboxed vectors mind you ... ), BUT that actually doesnt matter because any read/write is strict ish enough, and afaik, the only reason to provide the RNF instances is to reduce downstream boilerplate pain for deriving deepSeq instances for composite data structures.

What i mean is: unboxed vectors are philosophically a sort of n-ary tree of product types that bottoms out into some set of ground Prim Vector instances. So we *could* do some sort of generic deriving thing for RNF definitions for unboxed vectors that recurs rnf down to the leaves, which would be sort of @RyanGlScott 's wheel house, but I think that its not needed, though I could be wrong

cartazio:
examples of some vector layout families that are spiritually equivalent to unboxed which make the tree structure explicit are 

https://github.com/wellposed/numerical/blob/06655ba0f799b0548636e9977057f5bd1bc1caab/src/Numerical/Data/Vector/HPair.hs#L61-L91

and https://github.com/wellposed/numerical/blob/06655ba0f799b0548636e9977057f5bd1bc1caab/src/Numerical/Data/Vector/Pair.hs#L34-L62

in my numerical code base

PR #235, opened 2018-12-30
 Make Storable vectors nominally roled, and add unsafeCoerceVector functions

RyanGlScott:
In #224, we came to the conclusion that `Storable` vectors should be nominally roled in their type arguments. This PR implements that. I've also added `unsafeCoerce{M}Vector` functions which allow one to `coerce` between two `Storable` vectors in the event where you (the programmer) can ensure that the two `Storable` instances you're coercing between have identical memory representations.

There's still an open question about what to name these `unsafeCoerce{M}Vector` functions—suggestions welcome.

### DIFF
--- a/Data/Vector/Storable.hs
+++ b/Data/Vector/Storable.hs
@@ -132,7 +132,7 @@ module Data.Vector.Storable (
   toList, fromList, fromListN,
 
   -- ** Other vector types
-  G.convert, unsafeCast,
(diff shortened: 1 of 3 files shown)


cartazio:
@RyanGlScott do we still need to do this change in master?

RyanGlScott:
Yes. Also, opinions about what to name these functions are welcome.

PR #238, opened 2019-03-23
Adapt to MonadFail-related changes in base-4.13

RyanGlScott:
`base-4.13` will finally complete the `MonadFail` proposal by removing the `fail` method from `Monad`. This adapts `vector` to that change by guarding the existing `fail` implementation in `vector` (for `Data.Vector.Vector`) behind CPP, and defining a `MonadFail` instance for `Vector` going forward.

### DIFF
--- a/Data/Vector.hs
+++ b/Data/Vector.hs
@@ -175,9 +175,10 @@ import Control.DeepSeq ( NFData(rnf)
 #endif
                        )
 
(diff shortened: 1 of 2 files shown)


cartazio:
Thx.  I’ll review promptly

andrewthad:
I've reviewed this, and it works as intended. Merging.

cartazio:
the patch was inccorect and shouldn't have been merged as is :(

PR #243, opened 2019-05-01
Support GHC 8.8

fumieval:
Pardon an automatically-posted pull request; this is a part of [Operation Vanguard](https://github.com/haskell-vanguard/haskell-vanguard). Note that this does not guarantee that tests and benchmarks are buildable.

### DIFF
--- a/vector.cabal
+++ b/vector.cabal
@@ -147,13 +147,13 @@ Library
   Install-Includes:
         vector.h
 
-  Build-Depends: base >= 4.5 && < 4.13
+  Build-Depends: base >= 4.5 && < 4.14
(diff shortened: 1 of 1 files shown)


RyanGlScott:
The `base` version bump has already occurred on `master` (and on Hackage). The only remaining change relates to `semigroups-0.19`, which is unrelated to GHC 8.8.

chessai:
Is the release of `semigroups` tied to ghc releases, as `base` is? I wasn't aware if so. If not, then why make this change?

fumieval:
Looks like I was late by a day, while I was checking out too new semigroups.

PR #247, opened 2019-05-29
Add argument haddock for `copy` and `move`

minoki:
This is a complement to #190.

Adds `target`/`source` annotations to

- `Data.Vector.Mutable.copy`/`move`
- `Data.Vector.Primitive.Mutable.move`
- `Data.Vector.Storable.Mutable.move`.


RyanGlScott:
Thanks!

PR #270, opened 2020-01-29
Fix crash in specialized enumFromTo

### LINKED ISSUE #188: Segfault when using enumFromTo with maxBound as upper bound
I get a segfault when running the following code:

```Haskell
import Data.Word

main :: IO ()
main = (V.enumFromTo (pred maxBound) maxBound :: V.Vector Word8) `seq` return ()
```

```
$ cabal --version
cabal-install version 2.0.0.0
compiled using version 2.0.0.2 of the Cabal library 
$ ghc --version
The Glorious Glasgow Haskell Compilation System, version 8.2.1
```

I now use `enumFromN` as a workaround, that works.

This happens with `vector-0.12.0.1` from hackage.

Shimuuar:
This is attempt to fix #188. I had to change stream state from `a` to `Maybe a` in order to mark end of the stream. I also added regression tests in case someone will try to optimize these functions in the future

Fix is only partial. `Data.Vector.Fusion.Stream.Monadic` is affected by exactly same problem but it's not fixed yet

### DIFF
--- a/Data/Vector/Fusion/Bundle/Monadic.hs
+++ b/Data/Vector/Fusion/Bundle/Monadic.hs
@@ -758,13 +758,15 @@ enumFromTo x y = fromList [x .. y]
 -- FIXME: add "too large" test for Int
 enumFromTo_small :: (Integral a, Monad m) => a -> a -> Bundle m v a
 {-# INLINE_FUSED enumFromTo_small #-}
-enumFromTo_small x y = x `seq` y `seq` fromStream (Stream step x) (Exact n)
+enumFromTo_small x y = x `seq` y `seq` fromStream (Stream step (Just x)) (Exact n)
   where
     n = delay_inline max (fromIntegral y - fromIntegral x + 1) 0
 
     {-# INLINE_INNER step #-}
-    step z | z <= y    = return $ Yield z (z+1)
-           | otherwise = return $ Done
+    step Nothing              = return $ Done
+    step (Just z) | z == y    = return $ Yield z Nothing
+                  | z <  y    = return $ Yield z (Just (z+1))
+                  | otherwise = return $ Done
 
 {-# RULES
 
@@ -808,7 +810,7 @@ enumFromTo_small x y = x `seq` y `seq` fromStream (Stream step x) (Exact n)
 
 enumFromTo_int :: forall m v. Monad m => Int -> Int -> Bundle m v Int
 {-# INLINE_FUSED enumFromTo_int #-}
-enumFromTo_int x y = x `seq` y `seq` fromStream (Stream step x) (Exact (len x y))
+enumFromTo_int x y = x `seq` y `seq` fromStream (Stream step (Just x)) (Exact (len x y))
   where
     {-# INLINE [0] len #-}
     len :: Int -> Int -> Int
@@ -820,12 +822,14 @@ enumFromTo_int x y = x `seq` y `seq` fromStream (Stream step x) (Exact (len x y)
         n = v-u+1
 
     {-# INLINE_INNER step #-}
-    step z | z <= y    = return $ Yield z (z+1)
-           | otherwise = return $ Done
+    step Nothing              = return $ Done
+    step (Just z) | z == y    = return $ Yield z Nothing
+                  | z <  y    = return $ Yield z (Just (z+1))
+                  | otherwise = return $ Done
 
 enumFromTo_intlike :: (Integral a, Monad m) => a -> a -> Bundle m v a
 {-# INLINE_FUSED enumFromTo_intlike #-}
-enumFromTo_intlike x y = x `seq` y `seq` fromStream (Stream step x) (Exact (len x y))
+enumFromTo_intlike x y = x `seq` y `seq` fromStream (Stream step (Just x)) (Exact (len x y))
   where
     {-# INLINE [0] len #-}
     len u v | u > v     = 0
@@ -836,8 +840,10 @@ enumFromTo_intlike x y = x `seq` y `seq` fromStream (Stream step x) (Exact (len 
         n = v-u+1
 
     {-# INLINE_INNER step #-}
-    step z | z <= y    = return $ Yield z (z+1)
-           | otherwise = return $ Done
+    step Nothing              = return $ Done
+    step (Just z) | z == y    = return $ Yield z Nothing
+                  | z <  y    = return $ Yield z (Just (z+1))
+                  | otherwise = return $ Done
 
 {-# RULES
 
@@ -860,7 +866,7 @@ enumFromTo_intlike x y = x `seq` y `seq` fromStream (Stream step x) (Exact (len 
 
 enumFromTo_big_word :: (Integral a, Monad m) => a -> a -> Bundle m v a
 {-# INLINE_FUSED enumFromTo_big_word #-}
-enumFromTo_big_word x y = x `seq` y `seq` fromStream (Stream step x) (Exact (len x y))
+enumFromTo_big_word x y = x `seq` y `seq` fromStream (Stream step (Just x)) (Exact (len x y))
   where
     {-# INLINE [0] len #-}
     len u v | u > v     = 0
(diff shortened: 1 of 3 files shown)


cartazio:
Lovely!  I’ll take a look later today.

cartazio:
@Shimuuar  : https://travis-ci.org/haskell/vector/jobs/643472246#L653
looks like you need to depeend on the userland proxy package for older versions of ghc

cartazio:
(which i believe is just a reexport on newer)

Shimuuar:
Or just define proxy locally, it takes 1 line

lehins:
Or use `Maybe` ;P

instead of `typeRep (Proxy :: Proxy Int)` do `typeRep (Nothing :: Maybe Int)`

Shimuuar:
Code golf season is open `[] :: [Int]`, also `typeRep` doesn't exists on sufficiently old GHC (which vector supports)

lehins:
Yeah, I was gonna suggest using `typeRep Proxy`, instead of `typeOf undefined`, but then remembered that former isn't old enough. (I used it in a comment just to demonstrate proxy alternative)

cartazio:
I do like the p x, approach that allows empty list or proxy or maybe etc.

Shimuuar:
I think this PR is now ready

lehins:
I wonder if this fix affects the performance or ghc is smart enough to optimize the `Maybe` away. Would be interesting to see a benchmark (not suggesting that it is a requirement though):

An alternative solution would be to use the `Bounded` information (we can do that cause we know ahead of time the types this functions will be used with)
```haskell
enumFromTo_intlike :: (Bounded a, Integral a, Monad m) => a -> a -> Stream m a
{-# INLINE_FUSED enumFromTo_intlike #-}
enumFromTo_intlike x y =
  x `seq`
    if y == maxBound
      then Stream stepMaxBound (Just x)
      else Stream step x
  where
    {-# INLINE_INNER step #-}
    step z | z <= y    = return $ Yield z (z+1)
           | otherwise = return $ Done
    {-# INLINE_INNER stepMaxBound #-}
    stepMaxBound Nothing              = return $ Done
    stepMaxBound (Just z) | z == y    = return $ Yield z Nothing
                  | z <  y    = return $ Yield z (Just (z+1))
                  | otherwise = return $ Done
```

@Shimuuar any thoughts?

Shimuuar:
Benchmarks are absolutely necessary! Without benchmarks we could guide
optimization with augury

Another possibility is to add another constructor YieldLast to step data
type. But this is much bigger change

Data/Vector/Fusion/Bundle/Monadic.hs
```diff
@@ -758,13 +758,15 @@ enumFromTo x y = fromList [x .. y]
 -- FIXME: add "too large" test for Int
 enumFromTo_small :: (Integral a, Monad m) => a -> a -> Bundle m v a
 {-# INLINE_FUSED enumFromTo_small #-}
-enumFromTo_small x y = x `seq` y `seq` fromStream (Stream step x) (Exact n)
+enumFromTo_small x y = x `seq` y `seq` fromStream (Stream step (Just x)) (Exact n)
```
lehins: Any reason why we can't do this instead?

```haskell
enumFromTo_small :: (Integral a, Monad m) => a -> a -> Bundle m v a
{-# INLINE_FUSED enumFromTo_small #-}
enumFromTo_small x y = x `seq` y `seq` fromStream (S.enumFromTo_small x y) (Exact n)
  where
    n = delay_inline max (fromIntegral y - fromIntegral x + 1) 0
```

Data/Vector/Fusion/Bundle/Monadic.hs
```diff
@@ -808,7 +810,7 @@ enumFromTo_small x y = x `seq` y `seq` fromStream (Stream step x) (Exact n)
 
 enumFromTo_int :: forall m v. Monad m => Int -> Int -> Bundle m v Int
 {-# INLINE_FUSED enumFromTo_int #-}
-enumFromTo_int x y = x `seq` y `seq` fromStream (Stream step x) (Exact (len x y))
+enumFromTo_int x y = x `seq` y `seq` fromStream (Stream step (Just x)) (Exact (len x y))
```
lehins: Same here.

```haskell
enumFromTo_int :: forall m v. Monad m => Int -> Int -> Bundle m v Int
{-# INLINE_FUSED enumFromTo_int #-}
enumFromTo_int x y = x `seq` y `seq` fromStream (S.enumFromTo_int x y) (Exact (len x y))
...
```

Data/Vector/Fusion/Bundle/Monadic.hs
```diff
@@ -820,12 +822,14 @@ enumFromTo_int x y = x `seq` y `seq` fromStream (Stream step x) (Exact (len x y)
...
-           | otherwise = return $ Done
+    step Nothing              = return $ Done
+    step (Just z) | z == y    = return $ Yield z Nothing
+                  | z <  y    = return $ Yield z (Just (z+1))
+                  | otherwise = return $ Done
 
 enumFromTo_intlike :: (Integral a, Monad m) => a -> a -> Bundle m v a
 {-# INLINE_FUSED enumFromTo_intlike #-}
-enumFromTo_intlike x y = x `seq` y `seq` fromStream (Stream step x) (Exact (len x y))
+enumFromTo_intlike x y = x `seq` y `seq` fromStream (Stream step (Just x)) (Exact (len x y))
```
lehins: And here:

```haskell
enumFromTo_intlike :: (Integral a, Monad m) => a -> a -> Bundle m v a
{-# INLINE_FUSED enumFromTo_intlike #-}
enumFromTo_intlike x y = x `seq` y `seq` fromStream (S.enumFromTo_intlike x y) (Exact (len x y))
```

Data/Vector/Fusion/Bundle/Monadic.hs
```diff
@@ -860,7 +866,7 @@ enumFromTo_intlike x y = x `seq` y `seq` fromStream (Stream step x) (Exact (len
 
 enumFromTo_big_word :: (Integral a, Monad m) => a -> a -> Bundle m v a
 {-# INLINE_FUSED enumFromTo_big_word #-}
-enumFromTo_big_word x y = x `seq` y `seq` fromStream (Stream step x) (Exact (len x y))
+enumFromTo_big_word x y = x `seq` y `seq` fromStream (Stream step (Just x)) (Exact (len x y))
```
lehins: Here too:
```haskell
enumFromTo_big_word :: (Integral a, Monad m) => a -> a -> Bundle m v a
{-# INLINE_FUSED enumFromTo_big_word #-}
enumFromTo_big_word x y = x `seq` y `seq` fromStream (S.enumFromTo_big_int x y) (Exact (len x y))
```

Data/Vector/Fusion/Bundle/Monadic.hs
```diff
@@ -901,7 +909,7 @@ enumFromTo_big_word x y = x `seq` y `seq` fromStream (Stream step x) (Exact (len
 -- FIXME: the "too large" test is totally wrong
 enumFromTo_big_int :: (Integral a, Monad m) => a -> a -> Bundle m v a
 {-# INLINE_FUSED enumFromTo_big_int #-}
-enumFromTo_big_int x y = x `seq` y `seq` fromStream (Stream step x) (Exact (len x y))
+enumFromTo_big_int x y = x `seq` y `seq` fromStream (Stream step (Just x)) (Exact (len x y))
```
lehins: And here:
```haskell
enumFromTo_big_int :: (Integral a, Monad m) => a -> a -> Bundle m v a
{-# INLINE_FUSED enumFromTo_big_int #-}
enumFromTo_big_int x y = x `seq` y `seq` fromStream (S.enumFromTo_big_int x y) (Exact (len x y))
```

lehins:
@Shimuuar I'll experiment with suggestions I was making. For now, as you said, let's get the bug fixed (which is what implemented in this PR) and worry about improvements later.

So, @cartazio LGTM to be merged.

cartazio:
could you rebase it ontop of current master so the CI can be green? (i'm sure its fine, but better to be paranoid and make sure failures are failures )

cartazio:
eh, i'll do a local thingy

cartazio:
oh, you did a rebase and ci is done before i even could do the local thingy :)

cartazio:
dumb question: before i merge,  why does your commit say it doesn't fix Stream when it looks like it does? (is that just a stale commit message? If so i'll clean it up in my squash merge)

cartazio:
this is gonna need a rebase / conflict resolutlion, bleh

cartazio:
i'm gonna push a rebase branch for ya'll to sanity check then i'll apply it to master as is

cartazio:
cherrypicked, conflicts resolved, and merged

cartazio:
its in master!

cartazio:
(typos and all )

Issue #268, opened 2020-01-21
enumFromTo fails QC at -O2


paniag:
`$ ghc --version`
`The Glorious Glasgow Haskell Compilation System, version 8.6.5`
`$ cabal --version`
`cabal-install version 3.0.0.0`
`compiled using version 3.0.0.0 of the Cabal library`

Repro:
`$ cabal get vector`
`$ cd vector-0.12.0.3`
`$ echo "packages: vector.cabal" > cabal.project`
`$ cabal configure --disable-benchmarks --enable-tests && cabal install --only-dep`
`$ cabal test`

[vector-0.12.0.3-vector-tests-O2.log](https://github.com/haskell/vector/files/4088935/vector-0.12.0.3-vector-tests-O2.log)

cartazio:
thanks! 
Ok, it looks like the vector specialization/enumfromto rule for Double is wrong! eeek

cartazio:
@amosr eek, the bad enumfromto strikes again!

cartazio:
@nomeata fyi: nasty rules bug living here :)

paniag:
O0 passed (chance?)
Running O1 now.

[vector-0.12.0.3-vector-tests-O0.log](https://github.com/haskell/vector/files/4088949/vector-0.12.0.3-vector-tests-O0.log)

cartazio:
```
module Main where

import qualified MyLib (someFunc)
import qualified Data.Vector as V
main :: IO ()
main = do
  print ( V.fromList (enumFromTo 6.511955808615705e-2 18.451281790142502) == (V.enumFromTo 6.511955808615705e-2 18.451281790142502) )
 ```
outputs false on O1 :(

**Comment by cartazio** [comment]
based upon the output from  ` ghc-options: -O1  -ddump-rule-firings`
i'm going to say 
```
    enumFromTo<Double> [Bundle] (Data.Vector.Fusion.Bundle.Monadic)
```
is the culprit. I'm not sure why this wasn't exposed before in the CI ...

paniag:
-O1 predictably failed too.

[vector-0.12.0.3-vector-tests-O1.log](https://github.com/haskell/vector/files/4089002/vector-0.12.0.3-vector-tests-O1.log)

Shimuuar:
#188 is certainly related

Shimuuar:
I dug a little deeper. Failure is due to some subtle difference in how enumFromTo is implemented.

```haskell
import qualified Data.Vector as V
[...]

PR #271, opened 2020-01-29
Fix test failure for #268

Shimuuar:
GHC changed defition of Enum instance for Double in 8.6 release so we have to
accomodate both variants of enumFromTo

Draft for same reason as #270 identical bug in Stream module is not fixed yet

### DIFF
--- a/Data/Vector/Fusion/Bundle/Monadic.hs
+++ b/Data/Vector/Fusion/Bundle/Monadic.hs
@@ -952,7 +952,7 @@ enumFromTo_char x y = x `seq` y `seq` fromStream (Stream step xn) (Exact n)
 
 enumFromTo_double :: (Monad m, Ord a, RealFrac a) => a -> a -> Bundle m v a
 {-# INLINE_FUSED enumFromTo_double #-}
-enumFromTo_double n m = n `seq` m `seq` fromStream (Stream step n) (Max (len n lim))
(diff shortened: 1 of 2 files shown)


Shimuuar:
I tested with GHC 8.4 & 8.6. Both work

cartazio:
@Shimuuar CI should now be able to cover all the versions in question, many thanks to @lehins  for that patch

cartazio:
@Shimuuar ! thanks, i'm going to talk with GHC folks about this. this change didn't get reviewd by the core libraries committee in base, and it did change the semantics of float ops in base.

Shimuuar:
I think this PR is now ready

cartazio:
@Shimuuar dumb question, does the definition also agree with base when run on Floats as well as Doubles?

Shimuuar:
I think they should since tests pass. I dont remember whether there's rule
flotas too

чт, 30 янв. 2020 г., 23:43 Carter Tazio Schonwald <notifications@github.com

cartazio:
theres both, i added a test and both pass

cartazio:
right below where you wrote your patch :) 
```haskell 
"enumFromTo<Double> [Bundle]"
  enumFromTo = enumFromTo_double :: Monad m => Double -> Double -> Bundle m v Double

"enumFromTo<Float> [Bundle]"
  enumFromTo = enumFromTo_double :: Monad m => Float -> Float -> Bundle m v Float   #-}

```

PR #273, opened 2020-01-29
Split up the properties in the test suite into submodules: 

### LINKED ISSUE #272: Building test suite takes way too much memory
Building tests take a long time which is quite annoying for developing. But what even worse it uses a _lot_ of memory: 17GB with GHC8.6 & 14GB with GHC8.4. That's clearly a problem for anyone trying to work on vector on laptop. 

I don't know how bring memory consumption down. Maybe split Tests.Vector?

lehins:
Fixes #272

* Split happens on the type of vector
* Discovered that only inlining of Unboxed vector functions is affected.
* Disable inlining on `testPolymorphicFunctions` for Unboxed to workaround
  the blowup.

This PR reduces memory consumption during building the test suite to ~1GB

CC @cartazio


tests/Tests/Vector.hs
```diff
@@ -1,737 +1,15 @@
...
--- copied from GHC source code
-partitionWith :: (a -> Either b c) -> [a] -> ([b], [c])
-partitionWith _ [] = ([],[])
-partitionWith f (x:xs) = case f x of
-                         Left  b -> (b:bs, cs)
-                         Right c -> (bs, c:cs)
-    where (bs,cs) = partitionWith f xs
-
-testTuplyFunctions:: forall a v. (CommonContext a v, VectorContext (a, a) v, VectorContext (a, a, a) v) => v a -> [Test]
-testTuplyFunctions _ = $(testProperties [ 'prop_zip, 'prop_zip3
```
cartazio: why did you disable mzip funcitons?

lehins: I didn't disable them, I moved them over to `Monad` test, since those can only be run for boxed vector

tests/Tests/Vector/Property.hs
```diff
@@ -0,0 +1,647 @@
...
+
+testMonadFunctions :: forall a v. (CommonContext a v, VectorContext (a, a) v, MonadZip v) => v a -> [Test]
+{-# INLINE testMonadFunctions #-}
+testMonadFunctions _ = $(testProperties [ 'prop_return, 'prop_bind
+                                        , 'prop_mzip, 'prop_munzip])
+  where
+    prop_return :: P (a -> v a) = return `eq` return
+    prop_bind   :: P (v a -> (a -> v a) -> v a) = (>>=) `eq` (>>=)
+    prop_mzip   :: P (v a -> v a -> v (a, a)) = mzip `eq` zip
+    prop_munzip :: P (v (a, a) -> (v a, v a)) = munzip `eq` unzip
```
lehins: @cartazio See these `prop_mzip` and `prop_munzip`. Same properties as before, but they are only now applicable to any vector that is a monad (or rather `MonadFix`), and only boxed vector falls in into that category

cartazio: OK! :)

tests/Tests/Vector.hs
lehins: You see, they are restricted to `Data.Vector.Vector a`, instead of `v a`:
```
    prop_mzip   :: P (Data.Vector.Vector a -> Data.Vector.Vector a -> Data.Vector.Vector (a, a))
        = mzip `eq` zip
    prop_munzip :: P (Data.Vector.Vector (a, a) -> (Data.Vector.Vector a, Data.Vector.Vector a))
        = munzip `eq` unzip
```

which is sort dumb, cause basically it ignores the actual type of vector this `testTuplyFunctions` is currently being used for

cartazio:
@lehins  :))))

cartazio:
now to make things green with the rules patches

GeorgeCo:
great news and progress!

PR #279, opened 2020-01-31
Fix issues with vector slice function

### LINKED ISSUE #257: Segfaults and "internal errors" when vector slice overflows
I reported this for GHC first. See <https://gitlab.haskell.org/ghc/ghc/issues/17233> for full details (could be copied here if clearly a bug in vector). May be related to #188?

lehins:
Fixes #257

### DIFF
--- a/Data/Vector/Generic.hs
+++ b/Data/Vector/Generic.hs
@@ -466,10 +466,12 @@ unsafeDrop :: Vector v a => Int -> v a -> v a
 {-# INLINE unsafeDrop #-}
 unsafeDrop n v = unsafeSlice n (length v - n) v
 
-{-# RULES
 
-"slice/new [Vector]" forall i n p.
-  slice i n (new p) = new (New.slice i n p)
+-- Turned off due to: https://github.com/haskell/vector/issues/257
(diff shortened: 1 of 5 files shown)


cartazio:
wonderful! of course we need to fix that fusion rule, but this is better than not. 

(i think we could maybe change the rules around to push around the interval information into the fused code, but thats not what we 're here for today)

lehins:
To be honest I am pretty sure it is possible fixing that fusion rule without changing semantics of `slice` not throwing an error.

cartazio:
it is possible or *impossible*? i'm confused :)

lehins:
Sorry, not possible :man_facepalming:

cartazio:
i'm going to say "yes, impossible if we dont change the rules/fusion stuff", but i THINK if we add some hooks for "evidence" passing that we can later erase, there might be a way :)

cartazio:
this may be slightly beyond what GHC can currently do, and that might motivate some fun slow motion compiler hacking

lehins:
I've been wrong before ;) There might be a trick that I am simply not aware of.

PR #285, opened 2020-01-31
Fix selection of the resulting size when `drop` is applied.

### LINKED ISSUE #282: Fusion of `take` can cause OutOfMemory
Here is a simple repro:

```haskell
import qualified Data.Vector.Primitive as V -- or any other for that matter

main = do
  print $ V.take (maxBound `div` 8) $ V.fromList xs
```
Results in out of memory:

```
$ stack exec -- ghc take.hs -O1 -fforce-recomp && ./take
[1 of 1] Compiling Main             ( take.hs, take.o )
Linking take ...
take: Out of memory
```

Same thing as it was observed with `slice` in #257 and fixed in #279, except I think we can fix this issue without giving up fusion of `take` function.

The actual cause for this is that after a few rewrite rules we get into a situation where a smaller size being selected: https://github.com/haskell/vector/blob/f0cf694ec6feaae97f08035eefcf0c7ec76931cc/Data/Vector/Fusion/Bundle/Monadic.hs#L269-L272

and between `Unknown` and `Exact maxBound` ironically the smaller one is `Max maxBound` :)

```haskell
>>> smaller (Exact maxBound) Unknown
Max 9223372036854775807
```

Which later simply causes the out of memory when `vmunstreamMax` tries to allocate too much.

_Side note_ - `V.take maxBound $ V.fromList xs` is guarded by `basicUnsafeNew` up to `div maxBound (sizeOf (undefined :: e))`, but not for smaller than than and no guards for boxed vector.

lehins:
Fixes #282 - Possibility of OutOfMemory when argument to `drop` is too large

### DIFF
--- a/Data/Vector/Fusion/Bundle/Monadic.hs
+++ b/Data/Vector/Fusion/Bundle/Monadic.hs
@@ -269,7 +269,7 @@ tail Bundle{sElems = s, sSize = sz} = fromStream (S.tail s) (sz-1)
 -- | The first @n@ elements
 take :: Monad m => Int -> Bundle m v a -> Bundle m v a
 {-# INLINE_FUSED take #-}
-take n Bundle{sElems = s, sSize = sz} = fromStream (S.take n s) (smaller (Exact n) sz)
+take n Bundle{sElems = s, sSize = sz} = fromStream (S.take n s) (smallerThan n sz)
 
 -- | All but the first @n@ elements
 drop :: Monad m => Int -> Bundle m v a -> Bundle m v a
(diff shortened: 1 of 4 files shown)


cartazio:
(i pushed an addition to changelog that conflicts, :)) )

cartazio:
also seriously great work, thanks

lehins:
Now inability to force push to a branch became a problem, since in order for me to fix this conflict I had to delete the branch thus closing this PR. Switching all my work to the fork. Here is new PR: #286

cartazio:
Yeah.  I talked with admins about it.
Sounds good .   Might be a better workflow from a ci perspective cause of
how Travis limits jobs per org/user rather than some other way.  (Eg when
there’s a lot of patches for cabal, vector has to wait a while to run)

PR #286, opened 2020-01-31
Fix selection of the resulting size when `take` is applied:

### LINKED ISSUE #282: Fusion of `take` can cause OutOfMemory

lehins:
* Fixes #282 - Possibility of OutOfMemory when argument to `take` is too
large

### DIFF
--- a/Data/Vector/Fusion/Bundle/Monadic.hs
+++ b/Data/Vector/Fusion/Bundle/Monadic.hs
@@ -269,7 +269,7 @@ tail Bundle{sElems = s, sSize = sz} = fromStream (S.tail s) (sz-1)
 -- | The first @n@ elements
 take :: Monad m => Int -> Bundle m v a -> Bundle m v a
 {-# INLINE_FUSED take #-}
-take n Bundle{sElems = s, sSize = sz} = fromStream (S.take n s) (smaller (Exact n) sz)
+take n Bundle{sElems = s, sSize = sz} = fromStream (S.take n s) (smallerThan n sz)
 
 -- | All but the first @n@ elements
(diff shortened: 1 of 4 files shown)


cartazio:
@lehins do we have the patch for drop merged in already? i'm trying to find the PR for it but i cant, though i see the commit mentioned somewhere.

lehins:
@cartazio originally mixed up `take` and  `drop` in commit message, but not in the actual fix which is in this PR

PR #288, opened 2020-02-01
Bring back `mkType` function and deprecate it. Fix #287 Also:

### LINKED ISSUE #287: 0.12.1.1 breaking change w.r.t. 0.12.0.3
It looks as if `vector-0.12.1.1` has renamed the `Vector.Generic.mkType` function to `Vector.Generic.mkVecType`.

cf. [`mkType` @ 0.12.0.3](https://github.com/haskell/vector/blob/19189a884434ecd2c96212395f5705347b55775b/Data/Vector/Generic.hs#L2195-L2197) versus [`mkVecType` @ 0.12.1.1](https://github.com/haskell/vector/blob/release/v0.12.1/Data/Vector/Generic.hs#L2208-L2210)

lehins:
* Bump up the version in cabal and update changelog.
* Fix markdown header formatting in changelog.

### DIFF
--- a/Data/Vector/Generic.hs
+++ b/Data/Vector/Generic.hs
@@ -165,7 +165,7 @@ module Data.Vector.Generic (
   liftShowsPrec, liftReadsPrec,
 
   -- ** @Data@ and @Typeable@
-  gfoldl, gunfold, dataCast, mkVecType, mkVecConstr
+  gfoldl, gunfold, dataCast, mkVecType, mkVecConstr, mkType
 ) where
 
 import           Data.Vector.Generic.Base
@@ -203,6 +203,8 @@ import Prelude hiding ( length, null,
 import qualified Text.Read as Read
 import qualified Data.List.NonEmpty as NonEmpty
 
+import qualified Data.Traversable as T (Traversable(mapM))
+
 #if __GLASGOW_HASKELL__ >= 707
 import Data.Typeable ( Typeable, gcast1 )
 #else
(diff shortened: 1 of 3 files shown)


Shimuuar:
Accorfing to PVP

> Deprecation. Deprecated entities (via a DEPRECATED pragma) SHOULD be counted as removed for the purposes of upgrading the API, because packages that use -Werror will be broken by the deprecation. In other words the new A.B SHOULD be greater than the previous A.B.

I think we should remove deprecation

chessai:
I just commented the same thing as @Shimuuar in the related issue. This violates PVP, unfortunately

lehins:
@chessai I sincerely do not agree with that part of PVP, but at the same time I really don't care about this `mkType` function. So here, dropped the `DEPRECATE` pragma

Data/Vector/Generic.hs
```diff
@@ -203,6 +203,8 @@ import Prelude hiding ( length, null,
 import qualified Text.Read as Read
 import qualified Data.List.NonEmpty as NonEmpty
 
+import qualified Data.Traversable as T (Traversable(mapM))
+
```
cartazio: was this shuffled  because of how you wanted to do the cpp to conditionalize mkNoRepType?

cartazio:
I think changing pvp stance on deprecation annotations is ultimately gated on improving how ghc handles those and related warnings. (as with many things, the real way to progress stuff turns into "contrib the nice things you want into ghc and friends")

cartazio:
@lehins  this patch looks good, i'd be inclined to undo the shuffle and have two cpp clause so import order styling doesn't get semi spurious reordering.

cartazio:
unrealtedly: 
https://gist.github.com/piscisaureus/3342247 is an example of how to improve the working with PR workflow by registering the PR refs with git for local manipulation

Data/Vector/Generic.hs
lehins: This was shuffled because function definitions should not appear between imports, if I didn't move it it would look like this for `base < 4.2.0`:
```haskell
import Data.Data ( Data, DataType, Constr, Fixity(Prefix),
                   mkDataType, mkConstr, constrIndex )
                   mkDataType, mkConstr, constrIndex,
                   mkNorepType )
mkNoRepType :: String -> DataType
mkNoRepType = mkNorepType

import qualified Data.Traversable as T (Traversable(mapM))
```

The slight improvement to the way I did `CPP` was to reduce multiple imports from the same module, but it had nothing to do with `Traversable` import. `hlint` is a great tool that points those flaws to you.

lehins:
Reverting the ordering of imports is a wrong decision here. See my answer to your comment.

Issue #213, opened 2018-05-15
Documentation issues in Data.Vector.Mutable


treeowl:
`MVector` is exported concretely, but its three fields are undocumented. In particular, it's not obvious to users whether the second `Int` is an offset or a length.

`slice` doesn't document what its `Int` arguments mean. `Data.Vector.Generic.basicUnsafeSlice` lacks that documentation as well. The explanation should be copied from `Data.Vector` in each case.

`init`, `tail`, `take`, `drop`, and `splitAt` have no documentation whatsoever, and in particular fail to note that they don't copy the underlying vector. The "unsafe" versions are also completely undocumented.

`unsafeNew` claims that the vector isn't initialized, but it actually is. Only *unboxed* vectors can be created uninitialized. The only difference between `unsafeNew` and `new` is that `new` checks that the requested size is sensible.

sjakobi:
Thanks for the hint regarding `slice`! I had hoped that I could find info on this on the issue tracker! :)

cartazio:
Thanks for tickling this ticket. These definitely should be improved. !

PR #289, opened 2020-02-01
Add MINIMAL pragma to Vector & MVector type classes

### LINKED ISSUE #11: Add Minimal annotations to Vector typeclasses
the core type classes seem to lack the minimal annotations that ghc 7.8 now supports. Seems worth adding! I'll try to make some time to add that this weekend

### LINKED ISSUE #14: Don't include fromList in Show instance
I think we shouldn't include `fromList` in the output of the `Show` instance and instead mirror the output for lists. Reasons:
- With the new `OverloadedLists` extension the list literal syntax is valid for creating vectors.
- The `fromList` part exists to allow programmers to copy the output back into the REPL to create a new vector. This doesn't actually work as the modules are typically imported qualified.
- The `fromList` part is noise that makes the output harder to read e.g. when two vectors are nested.

Shimuuar:
At the moment only checked that MINIMAL is indeed minimal and no method is ⊥ by visual inspection

Fixes #11

### DIFF
--- a/Data/Vector/Generic/Mutable.hs
+++ b/Data/Vector/Generic/Mutable.hs
@@ -595,8 +595,10 @@ new n = BOUNDS_CHECK(checkLength) "new" n
 --   should be presumed uninitialized. However exact semantics depends
 --   on vector implementation. For example unboxed and storable
 --   vectors will create vector filled with whatever underlying memory
---   buffer happens to contain, while boxed vector initializes every
---   element to @error "..."@.
+--   buffer happens to contain, while boxed vector's elements are
+--   initialized to bottoms which will throw exception when evaluated.
+--
+-- @since 0.4
 unsafeNew :: (PrimMonad m, MVector v a) => Int -> m (v (PrimState m) a)
 {-# INLINE unsafeNew #-}
 unsafeNew n = UNSAFE_CHECK(checkLength) "unsafeNew" n
--- a/Data/Vector/Generic/Mutable.hs
+++ b/Data/Vector/Generic/Mutable.hs
@@ -591,7 +591,12 @@ new :: (PrimMonad m, MVector v a) => Int -> m (v (PrimState m) a)
 new n = BOUNDS_CHECK(checkLength) "new" n
       $ unsafeNew n >>= \v -> basicInitialize v >> return v
 
--- | Create a mutable vector of the given length. The memory is not initialized.
+-- | Create a mutable vector of the given length. The vector content
+--   should be presumed uninitialized. However exact semantics depends
+--   on vector implementation. For example unboxed and storable
+--   vectors will create vector filled with whatever underlying memory
+--   buffer happens to contain, while boxed vector initializes every
+--   element to @error "..."@.
 unsafeNew :: (PrimMonad m, MVector v a) => Int -> m (v (PrimState m) a)
 {-# INLINE unsafeNew #-}
 unsafeNew n = UNSAFE_CHECK(checkLength) "unsafeNew" n
--- a/Data/Vector/Mutable.hs
+++ b/Data/Vector/Mutable.hs
@@ -281,12 +281,10 @@ new :: PrimMonad m => Int -> m (MVector (PrimState m) a)
 {-# INLINE new #-}
 new = G.new
 
--- | Create a mutable vector of the given length. The vector content
---   should be presumed uninitialized. However exact semantics depends
---   on vector implementation. For example unboxed and storable
---   vectors will create vector filled with whatever underlying memory
---   buffer happens to contain, while boxed vector initializes every
---   element to @error "..."@.
+-- | Create a mutable vector of the given length. The vector elements
+--   are set to bottom so accessing them will cause an exception.
+--
+-- @since 0.5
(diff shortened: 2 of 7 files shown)


Shimuuar:
Yes, list is indeed minimal. I also thrown in fix for #213

Data/Vector/Primitive/Mutable.hs
```diff
@@ -226,7 +226,12 @@ new :: (PrimMonad m, Prim a) => Int -> m (MVector (PrimState m) a)
 {-# INLINE new #-}
 new = G.new
 
--- | Create a mutable vector of the given length. The memory is not initialized.
+-- | Create a mutable vector of the given length. The vector content
+--   should be presumed uninitialized. However exact semantics depends
+--   on vector implementation. For example unboxed and storable
+--   vectors will create vector filled with whatever underlying memory
+--   buffer happens to contain, while boxed vector initializes every
+--   element to @error "..."@.
```
lehins, line 234:
```suggestion
-- | Create a mutable vector of the given length. The vector content
--   is uninitialized, which means it is filled with whatever underlying memory
--   buffer happens to contain.
--
-- @since 0.5
```

It is best to describe exact semantics, rather than all possible but unrelated variations. The original comment for generic implementation is great though.

Data/Vector/Storable/Mutable.hs
```diff
@@ -375,7 +375,12 @@ new :: (PrimMonad m, Storable a) => Int -> m (MVector (PrimState m) a)
 {-# INLINE new #-}
 new = G.new
 
--- | Create a mutable vector of the given length. The memory is not initialized.
+-- | Create a mutable vector of the given length. The vector content
+--   should be presumed uninitialized. However exact semantics depends
+--   on vector implementation. For example unboxed and storable
+--   vectors will create vector filled with whatever underlying memory
+--   buffer happens to contain, while boxed vector initializes every
+--   element to @error "..."@.
```
lehins, line 383:
```suggestion
-- | Create a mutable vector of the given length. The vector content
--   is uninitialized, which means it is filled with whatever underlying memory
--   buffer happens to contain.
--
-- @since 0.5
```

Data/Vector/Unboxed/Mutable.hs
```diff
@@ -155,7 +155,12 @@ new :: (PrimMonad m, Unbox a) => Int -> m (MVector (PrimState m) a)
 {-# INLINE new #-}
 new = G.new
 
--- | Create a mutable vector of the given length. The memory is not initialized.
+-- | Create a mutable vector of the given length. The vector content
+--   should be presumed uninitialized. However exact semantics depends
+--   on vector implementation. For example unboxed and storable
+--   vectors will create vector filled with whatever underlying memory
+--   buffer happens to contain, while boxed vector initializes every
+--   element to @error "..."@.
```
lehins, line 163:
```suggestion
-- | Create a mutable vector of the given length. The vector content
--   is uninitialized, which means it is filled with whatever underlying memory
--   buffer happens to contain.
--
-- @since 0.5
```

Data/Vector/Mutable.hs
```diff
@@ -281,7 +281,12 @@ new :: PrimMonad m => Int -> m (MVector (PrimState m) a)
 {-# INLINE new #-}
 new = G.new
 
--- | Create a mutable vector of the given length. The memory is not initialized.
+-- | Create a mutable vector of the given length. The vector content
+--   should be presumed uninitialized. However exact semantics depends
+--   on vector implementation. For example unboxed and storable
+--   vectors will create vector filled with whatever underlying memory
+--   buffer happens to contain, while boxed vector initializes every
+--   element to @error "..."@.
```
lehins, line 289:
```suggestion
-- | Create a mutable vector of the given length. Boxed vector is
--  initialized with every element set to @error "..."@.
--
-- @since 0.5
```

Data/Vector/Generic/Mutable.hs
```diff
@@ -591,7 +591,12 @@ new :: (PrimMonad m, MVector v a) => Int -> m (v (PrimState m) a)
 new n = BOUNDS_CHECK(checkLength) "new" n
       $ unsafeNew n >>= \v -> basicInitialize v >> return v
 
--- | Create a mutable vector of the given length. The memory is not initialized.
+-- | Create a mutable vector of the given length. The vector content
+--   should be presumed uninitialized. However exact semantics depends
+--   on vector implementation. For example unboxed and storable
+--   vectors will create vector filled with whatever underlying memory
+--   buffer happens to contain, while boxed vector initializes every
+--   element to @error "..."@.
```
lehins, line 599:
```suggestion
--   element to @error "..."@.
--
-- @since 0.4
```

Bodigrim:
I'd probably say that elements of boxed vectors are initialized with bottoms, but do not overspecify whether it is `error` or `undefined` or another diverging computation.

Shimuuar:
@lehins Thanks for the suggestion. They definitely improved haddocks. And thanks for finding out those `@since` pragmas

@Bodigrim Yeah, good idea. I also specified that it's exception throwing bottom.

Issue #299, opened 2020-02-09
current assymmetries in exports for Vector wrt Boxed/Unboxed/Generic/Storable


cartazio:
i've some half scripted stuff i used to generate comparisons of exports and help with me getting the backpack stuff started

this comparison uses v0.12.1.2 (current most recent release)
stuff in Generic but not in boxed: (alphabetical order)
```
clone
cmp
cmpBy
concatNE
eq
eqBy


gfoldl
gunfold
liftReadsPrec
liftShowsPrec

new

readPrec
showsPrec

stream
streamR

unstream
unstreamR
```
(some stuff spuriously shows up as being different when its Reexported with G.whatever,i've removed those from the list i generated)

missing stuff in storable
```
ifoldM
ifoldM'
ifoldM'_
ifoldM_
imapM
imapM_
iscanl
iscanl'
iscanr
iscanr'
izipWithM
izipWithM_
new

sequence
sequence_
unsafeAccumulate  
unsafeUpdate  
       
unstream
unstreamR
unzip
unzip3
unzip4
unzip5
unzip6
update

zip
zip3
zip4
zip5
zip6
```

and then in unboxed

```
clone
cmp
cmpBy
concatNE
eq
eqBy
gfoldl
gunfold
iscanl
iscanl'
iscanr
iscanr'
sequence
sequence_
```
i might be skipping mentionign 1-2 here,
but if backpack actually materially helps us PROPERLY enforce consistent apis and making it way user to write new flavores of instances with good inference, etc etc

cc @merijn  @chessai @andrewthad @lehins @Shimuuar  fyi
i'm skipping mentoing 1-2 like `(un)stream(R)` and the `Data` class helpers on 1-2 of the lists, but this should have good coverage of all the current warts

cartazio:
afaict, the best full of examples exposition of backpack is ezyang's well written thesis. (for those who are curious)
https://github.com/ezyang/thesis/releases/tag/rev20170925

cartazio:
also i'll try to cleanup these helper scripts so that other folks can easily hack them into having a mostly mechanical way to compare large explicit export lists

cartazio:
1) I’ve not done this comparison for other modules, should I?
2) would other folks like me to take a few days to cleanup these compare scripts for easy use?

lehins:
[...]

PR #300, opened 2020-02-16
Export unstreamM from Data.Vector.Generic

### LINKED ISSUE #70: unstreamM is not exported
We are trying to write a `scanlM` and `scanrM` for `vector`, but to do so it seems that we need `unstreamM`, which exists in `Generic.hs` but isn't exported. Is there a reason why it isn't?

Shimuuar:
I don't think there's any good reason for not exporting this functions and its
lack caused problems for users (#70).

Fixes #70

### DIFF
--- a/Data/Vector/Generic.hs
+++ b/Data/Vector/Generic.hs
@@ -149,7 +149,7 @@ module Data.Vector.Generic (
   -- * Fusion support
 
   -- ** Conversion to/from Bundles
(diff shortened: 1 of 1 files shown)


cartazio:
Sounds good.  Can you both weigh in on if the lists in https://github.com/haskell/vector/issues/299
Either overlook something or inclue something they shouldn’t ?

Issue #111, opened 2016-03-19
Length doesn't seem to inline properly


gbaz:
As discussed here: http://stackoverflow.com/questions/36095925/why-vector-length-vector-replicate-n-0-is-not-fused

In simple test programs, `Data.Vector.Generic.length` seems to refuse to inline, for no apparent reason.

See #97 as perhaps related?

gbaz:
In the SO discussion, it appears to be because in the latest Vector, ghc marks length as a loop-breaker.

cartazio:
Huh, is that version of length not written in terms of length on stream /
bundle or as a worker wrapper?

takano-akio:
I believe #155 fixes this.

Shimuuar:
Finally fixed in #307

Issue #229, opened 2018-11-17
A test suite for fusion


nomeata:
The vector library puts in lots of effort to ensure that pipelines of vector functions fuse properly, but this is not really tested. In the similar fusion system in `text`, I found that [not all intended fusion actually happened](https://github.com/haskell/text/issues/202), and used [`inspection-testing`](http://hackage.haskell.org/package/inspection-testing) to comprehensively check for fusion.

I believe the vector library would benefit from this as well. I started working on such a test suite, and want to use this issue to discuss the approach.

nomeata:
Currently, this tests fusion only against `fromList `and `toList`, two “known good consumers/producers”. It might be more interesting to create random expressions and pipelines.

I could imagine a list with entries like 

    [ ('map, O ~> V ~> F ~> V)
    , ('reverse,  V ~> F)
    , ('backpermute, V ~> F ~> V)
    ]

that specify the fusion-behavior of the function’s arguments and return type (`F = fusing vector`, `V = non-fusing vector`, `O = other type`), and then lego-like putting these pieces together to create big pipelines (or even trees – some functions can fuse with multiple arguments) to create more interesting test cases.

And maybe we can put these annotations somehow into the haddocks of the function, to publicly document the fusion behaviour.

cartazio:
Ok, this is very cool! I’m in the middle of holidays and fam stuff but I’m
doing some prelim stuff to vector.  This is a great little bunch.  I’ll
carve some time soon to work through what you’ve put down.

cartazio:
i'll pull this back onto my todo list in the next month or so,

cartazio:
this should be done after this current release this winter

Shimuuar:
I'm copying @Bodigrim's comment https://github.com/haskell/vector/pull/389#issuecomment-846565591

> Fusion tests in `text` were generated using my fork of `inspection-testing`: https://github.com/Bodigrim/inspection-testing/tree/6b99b613b30fc2216f391db08b857a54352e4f48/examples/text-api 
It should be possible to adapt the generator to `vector`.

PR #307, opened 2020-06-03
Make 'length' fusible

### LINKED ISSUE #306: Data.Vector.Generic.length not fusible
Unfortunately, the function `Data.Vector.Generic.length` is not currently covered by stream fusion.  This is because  `length` is defined in terms of `Data.Vector.Generic.stream'`, which does not admit fusion rules at all.

As a result, the following simple code:
https://gist.github.com/gksato/417030a1f19f1f376dcb2fabaf66c10a#file-test-hs
*does* require a *big* allocation: 
https://gist.github.com/gksato/417030a1f19f1f376dcb2fabaf66c10a#file-testbefore-ddump-simpl

I'm soon gonna submit a pull request to fix this!

gksato:
This change resolves #306: it makes fusible `Data.Vector.Generic.length`, which currently can be inlined but doesn't fuse well. Before this revision, `length` uses `stream'`, which is non-fusible.
This revision replaces `stream'` with `stream`, being fusible, and at the same time substitute the mutually-recursive occurrences of `length` with `basicLength`.

Rationale:
The current definition was introduced in the commit a811a863f58dd1521893c251466e8d3c8bb457e2.
Prior to that commit, `length` could not even be inlined, because GHC elected it to be a loop-breaker due to the cyclic references among `stream`, `clone`, `length`, and `unsafeCopy`. This failure of inlining was reported as #97.

The commit a811a863f resolved this by replacing `stream` with `stream'` in the definition of `length`. The function `stream` refers to `clone` in its fusion rule, but `stream'` doesn't possess any rule. This cuts open the cycle of references and makes `length` inlinable.

However, since `stream'` doesn't possess any rule, defining
``` Haskell
length = Bundle.length . stream'
```
is all the same as setting
``` Haskell
length = basicLength
```
This prevents any stream fusion from happening.

This pull request resolves this problem by resetting `length` back to be `Bundle.length . stream`, and instead replacing the occurrences of `length` making cycles with `basicLength` (just *inlined the simplification result of the definition*). Now `length` is both inlinable and fusible.

The comparison test of the compilation of a simple program is in https://gist.github.com/gksato/417030a1f19f1f376dcb2fabaf66c10a .

Fixes: https://github.com/haskell/vector/issues/306
See also: https://github.com/haskell/vector/issues/97 https://github.com/haskell/vector/issues/111 https://github.com/haskell/vector/pull/155

### DIFF
--- a/Data/Vector/Generic.hs
+++ b/Data/Vector/Generic.hs
@@ -222,7 +222,7 @@ import qualified Data.Traversable as T (Traversable(mapM))
 -- | /O(1)/ Yield the length of the vector
 length :: Vector v a => v a -> Int
 {-# INLINE length #-}
-length = Bundle.length . stream'
+length = Bundle.length . stream
 
 -- | /O(1)/ Test whether a vector is empty
 null :: Vector v a => v a -> Bool
@@ -1986,7 +1986,7 @@ copy
   :: (PrimMonad m, Vector v a) => Mutable v (PrimState m) a -> v a -> m ()
 {-# INLINE copy #-}
 copy dst src = BOUNDS_CHECK(check) "copy" "length mismatch"
-                                          (M.length dst == length src)
+                                          (M.length dst == basicLength src)
              $ unsafeCopy dst src
 
 -- | /O(n)/ Copy an immutable vector into a mutable one. The two vectors must
@@ -1995,7 +1995,7 @@ unsafeCopy
   :: (PrimMonad m, Vector v a) => Mutable v (PrimState m) a -> v a -> m ()
 {-# INLINE unsafeCopy #-}
 unsafeCopy dst src = UNSAFE_CHECK(check) "unsafeCopy" "length mismatch"
-                                         (M.length dst == length src)
+                                         (M.length dst == basicLength src)
                    $ (dst `seq` src `seq` basicUnsafeCopy dst src)
 
 -- Conversions to/from Bundles
@@ -2004,13 +2004,7 @@ unsafeCopy dst src = UNSAFE_CHECK(check) "unsafeCopy" "length mismatch"
 -- | /O(1)/ Convert a vector to a 'Bundle'
 stream :: Vector v a => v a -> Bundle v a
 {-# INLINE_FUSED stream #-}
-stream v = stream' v
-
--- Same as 'stream', but can be used to avoid having a cycle in the dependency
--- graph of functions, which forces GHC to create a loop breaker.
-stream' :: Vector v a => v a -> Bundle v a
-{-# INLINE stream' #-}
-stream' v = Bundle.fromVector v
+stream v = Bundle.fromVector v
 
 {-
 stream v = v `seq` n `seq` (Bundle.unfoldr get 0 `Bundle.sized` Exact n)
@@ -2133,7 +2127,7 @@ clone :: Vector v a => v a -> New v a
 {-# INLINE_FUSED clone #-}
 clone v = v `seq` New.create (
   do
-    mv <- M.new (length v)
+    mv <- M.new (basicLength v)
     unsafeCopy mv v
     return mv)
 


cartazio:
Sounds very awesome!  I’m going to make sure I look at this soon , though just in case I miss anything I’ll ask a few other folks to have a look too! (Lots of chaos in nyc etc this week )

Cc @sjakobi @Bodigrim @emilypi

gksato:
Alright! Stay safe.

gksato:
I simplified the test: https://gist.github.com/gksato/157a98da55c643f15760720383419b4b

Shimuuar:
I finally go to review this PR. I digged through code and now I mildly confused. Does original loop `stream, clone, length, unsafeCopy` still exists? It seems `stream` was rewritten in meantime to use Bundle. 

@gksato did you try to just use `stream` for `length` definition? 

P.S. this just illustrates how badly we need tests for fusion (#229). So far our answer to question does it fuse is "hopes and prayers" which is clearly not enough

Bodigrim:
@gksato could you possibly submit a test for this, using [`inspection-testing`](http://hackage.haskell.org/package/inspection-testing)?

gksato:
@Shimuuar Now I get the situation well. Just rewriting back
``` Haskell
length = Bundle.length . stream
```
*does* work with **GHC 8.8.3**. However it doesn't work with **GHC 8.0.2**, since there is, *formally*, a loop among `length`, `clone`, `unsafeCopy`, `stream`:

``` Haskell
-- | /O(1)/ Yield the length of the vector
length :: Vector v a => v a -> Int
{-# INLINE length #-}
length = Bundle.length . stream

-- | /O(1)/ Convert a vector to a 'Bundle'
stream :: Vector v a => v a -> Bundle v a
{-# INLINE_FUSED stream #-}
stream v = Bundle.fromVector v

{-# RULES
...

"New.unstream/stream [Vector]" forall v.
  New.unstream (stream v) = clone v

...  #-}

-- | Convert a vector to an initialiser which, when run, produces a copy of
-- the vector.
clone :: Vector v a => v a -> New v a
{-# INLINE_FUSED clone #-}
clone v = v `seq` New.create (
  do
    mv <- M.new (length v)
    unsafeCopy mv v
    return mv)

-- | /O(n)/ Copy an immutable vector into a mutable one. The two vectors must
-- have the same length. This is not checked.
unsafeCopy
  :: (PrimMonad m, Vector v a) => Mutable v (PrimState m) a -> v a -> m ()
{-# INLINE unsafeCopy #-}
unsafeCopy dst src = UNSAFE_CHECK(check) "unsafeCopy" "length mismatch"
                                         (M.length dst == length src)
                   $ (dst `seq` src `seq` basicUnsafeCopy dst src)
```

Since no `New.unstream` appears at the inline sites of `length`, this loop is only formal and not actual; I guess that's why reverting a811a863f58dd1521893c251466e8d3c8bb457e2  worked with **GHC 8.8.3**. However older GHCs seemingly doesn't detect that the loop is not essential.

gksato:
@Bodigrim I'm skimming through [`inspection-testing`](http://hackage.haskell.org/package/inspection-testing) and #229. Wait a moment...

Shimuuar:
Copy/unsafeCopy/clone require vector to be materialized anyway. So we don't lose anything by switching to basicLength. LGTM

@gksato Could you please add info from https://github.com/haskell/vector/pull/307#issuecomment-643032883 to commit message?

Shimuuar:
@Bodigrim @lehins what do you think about this PR?

gksato:
> @gksato Could you please add info from [#307 (comment)](https://github.com/haskell/vector/pull/307#issuecomment-643032883) to commit message?

@Shimuuar Oops. I've forgotten to check this PR. Can I reword the commit and force-push? I'll not change the content.

Shimuuar:
Please do

gksato:
@Shimuuar I've done it!

Shimuuar:
Did you look into inspection-testing by chance? Adding even single test would be great contribution since adding more will become relative straightforward.

gksato:
In fact, I was just puzzled by tests in https://github.com/haskell/vector/pull/307#issuecomment-643032883, because it included `test_length`. I now noticed that `test_length` was just defined as a test and was not included in the test to be executed.

gksato:
Here's a fusion test for this pull request. I picked up @nomeata's work on #229 and modified a bit:

- https://github.com/gksato/vector/tree/fusion-tests
commit e9e4e8b15eeecbee277aeb3b15b0471a96dd1d10
on branch `gksato:fusion-tests`.
This is the simple rebase of nomeata:fusion-tests onto the parent of this pull request.
- https://github.com/gksato/vector/tree/fusion-tests-length-failure
commit 75cf13c330d71182752a1a66779df1aab14c9ebd
on branch `gksato:fusion-tests-length-failure`.
Added, to `gksato:fusion-tests`, the test for `length`. The test fails.
- https://github.com/gksato/vector/tree/fusion-tests-length-success
commit 0fec58b93a6910f96553ac52969986a0eb385b2e
on branch `gksato:fusion-tests-length-success`.
This is a merge of `gksato:fusion-tests-length-failure` and this PR.
The test passes.

I wasn't really sure it was OK to merge a big mass of tests with this pull request, I left it separate branches. If you know a better-looking way to present these to you, please let me know!

Shimuuar:
I've merged PR, thanks!

It would be better to submit tests as separate PR. It will probably require several iterations to get into working order and I don't think it makes sense to keep this PR unmerged. 

P.S. Didn't look at tests yet

PR #302, opened 2020-04-11
Forward port changes from 0.12.1 branch

Shimuuar:
What is ported:

 - Changelog updates
 - Transition from test-framework to tasty
 - mkType (#287) is restored but deprecated
 - New tests for maxIndex & Co. I had to change them because of change of behavior

PR #308, opened 2020-06-04
Wip/carter/forward port 0.12.1 tests into master

cartazio:
@Shimuuar  @lehins  heres a hopefully simpler forward port that should be largely the same as 
https://github.com/haskell/vector/pull/302, but which keeps the semigroup trick for describing the associativities of the maxBy and MinBy

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -42,6 +42,25 @@ TODO: should this be in the next release
      `Data.Vector.Storable{.Mutable}` to allow this (the onus is on the user
      to ensure that no `Storable` invariants are broken when using these
      functions).
+
+# Changes in version 0.12.1.2
+
+ * Fix for lost function `Data.Vector.Generic.mkType`: [#287](https://github.com/haskell/vector/issues/287)
+ * all the changes in 0.12.1.0 and 0.12.1.1
+
+# Changes in version 0.12.1.1 (deprecated)
+ * add semigrioups dep to test suite so CI actually runs again on GHC < 8
+
+# Changes in version 0.12.1.0 (deprecated)
+ * Fix integer overflows in specializations of Bundle/Stream enumFromTo on Integral types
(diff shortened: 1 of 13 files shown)


cartazio:
derp, property test fail :)

cartazio:
oh, found the issue (relates to the stronger property test that doesn't work well for Float/Double with the current QC generators for the same types)

cartazio:
https://github.com/nick8325/quickcheck/issues/295  --- i'm commenting out one test in QC test for now, though we could copy the fix by @phadej intree perhaps if needed , im open to either approach

cartazio:
@lehins @Shimuuar i'm sure you two have some good change requests for this, i look forward to the feedback :) 

many thanks

cartazio:
at the end this should all get squashed, i'm a bit derpy and making 1-2 flubs in cleaning this up locally

cartazio:
theres 1-2 things to still cleanup once i get the tests green, like cleaning up how i fixed changing the TestTree type aliases in the modules, but should be good overall I hope :)

lehins:
@cartazio I appreciate you trying to help with this, but this is no way to port back/forward existing changes. This looks like a totally separate PR. The way @Shimuuar did it in #302 is the right way.
If there are some changes you'd like to add on top of the changes ported forward, feel free to submit a PR just with those changes. Make sure not to create any branches on this repo.

cartazio:
I guess I understand. I hope we can learn to work together in a positive
way in the future. And again I’m sorry that I’ve been frustrating to
collaovrate with in the past

The semigroup test is there to validate that the semantics of maxby In
vector vs base have the desired meaning with some clarity on the left vs
right tie breaking.

I’ve asked a friend who you may be more comfortable with to see about
preparing that itch

PR #309, opened 2020-06-05
Forward port 0.12.1

lehins:
This PR supersedes the #302 

@Shimuuar I squashed two commits together where one was removing part of changelog 0c79a1a4338cf540399d17866796dcc0a06a7ddb and the one that was bringing it back 00ffc50f746acf98e0d5894f1fc7a5637acfd978 Feel free to merge it if you ok with this squash, since the rest of the stuff was ported back perfectly

What is ported:

 - Changelog updates
 - Transition from test-framework to tasty
 - mkType (#287) is restored but deprecated
 - New tests for maxIndex & Co. I had to change them because of change of behavior

### DIFF
--- a/Data/Vector/Generic.hs
+++ b/Data/Vector/Generic.hs
@@ -2210,7 +2210,7 @@ mkVecType name = mkDataType name [mkVecConstr name]
 
 mkType :: String -> DataType
 {-# INLINE mkType #-}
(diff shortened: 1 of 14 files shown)


Shimuuar:
Thanks! I just wanted to ask to look at #302. Squashing is of course fine. When transplanting commits one really wants to make small commits or else it would be impossible to undertand what is going on

Issue #262, opened 2019-12-15
Missing API master issue


merijn:
This issue is to document functions that are in *some* of the vector modules, but not others. I will update and add missing functions as I find them:

Not anywhere, but nice to have:
  - `iforM`
  - `iforM_`

Mutable modules should probably re-export `PrimState` and `PrimMonad` so it's easier to write `PrimMonad` generic code without adding the additional `primitive` dependency and an extra import.

`Data.Vector.Storable`:
  - `imapM`
  - `imapM_`

cartazio:
yay, thx! this sort of concrete actionable feedback is really helpful!

theres some tools somewhere for doing haddock or .hi file based api diffs,
something like that would be super helpful to sanity check cleanup vector
with

merijn:
Right now I'm just writing down everything I'm having to reimplement to get my stuff working >.>
A substantial number are already in `Data.Vector.Generic`, so I'm unsure why they aren't exported already.

merijn:
Various monadic folds:

- `foldM`
- `ifoldM`

I suppose only strict left folds are sensible here?

cartazio:
Not necessarily.  In terms of stream fusion yes, but for materialized
vectors both are valid.

Shimuuar:
This leaves only

> Mutable modules should probably re-export PrimState and PrimMonad so it's easier to write PrimMonad generic code without adding the additional primitive dependency and an extra import.

I think it's fair request and won't cause name clashes since module is expected to imported qualified

PR #311, opened 2020-06-06
Implement iforM and iforM_ (#262)

fumieval:
Added iforM and iforM_, indexed variants of `forM` and `forM_`

### DIFF
--- a/Data/Vector.hs
+++ b/Data/Vector.hs
@@ -98,6 +98,7 @@ module Data.Vector (
 
   -- ** Monadic mapping
   mapM, imapM, mapM_, imapM_, forM, forM_,
+  iforM, iforM_,
 
   -- ** Zipping
   zipWith, zipWith3, zipWith4, zipWith5, zipWith6,
@@ -1062,6 +1063,18 @@ forM = G.forM
 forM_ :: Monad m => Vector a -> (a -> m b) -> m ()
 {-# INLINE forM_ #-}
 forM_ = G.forM_
+
+-- | /O(n)/ Apply the monadic action to all elements of the vector and their indices, yielding a
+-- vector of results. Equivalent to 'flip' 'imapM'.
+iforM :: Monad m => Vector a -> (Int -> a -> m b) -> m (Vector b)
+{-# INLINE iforM #-}
+iforM = G.iforM
+
+-- | /O(n)/ Apply the monadic action to all elements of the vector and their indices and ignore the
+-- results. Equivalent to 'flip' 'imapM_'.
+iforM_ :: Monad m => Vector a -> (Int -> a -> m b) -> m ()
+{-# INLINE iforM_ #-}
+iforM_ = G.iforM_
 
 -- Zipping
 -- -------
--- a/Data/Vector/Generic.hs
+++ b/Data/Vector/Generic.hs
@@ -81,6 +81,7 @@ module Data.Vector.Generic (
 
   -- ** Monadic mapping
   mapM, imapM, mapM_, imapM_, forM, forM_,
+  iforM, iforM_,
 
   -- ** Zipping
   zipWith, zipWith3, zipWith4, zipWith5, zipWith6,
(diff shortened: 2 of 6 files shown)


Shimuuar:
LGTM 

Also I vaguely remember that GHC required fully saturated application to inline. So `iforM xs` won't be inlined since only one parameter is supplied. Consequently it won't be specialized and everything will be _slow_. I'll try to invent test case for that

fumieval:
Right, I copypasted the definition of `forM` and `forM_`, assuming that they are written this way deliberately. I can update them (including existing `forM`) easily

Shimuuar:
> LGTM 

Sorry. I take that back. :) You forgot `Data.Vector.Primitive`

fumieval:
Good catch!

Shimuuar:
Amended PR LGTM. @lehins @Bodigrim could any of you take a look it just in case. I already almost missed Primitive reexport

No it seems GHC is good with current definition. Toy benchmark sees no difference:

```haskell
import Criterion.Main

fun1 :: (Int -> IO ()) -> IO ()
{-# NOINLINE fun1 #-}
fun1 = V.forM_ (V.generate 4000 id)

fun2 :: (Int -> IO ()) -> IO ()
{-# NOINLINE fun2 #-}
fun2 = \f -> V.forM_ (V.generate 4000 id) f

main :: IO ()
main = defaultMain
  [ bench "1" $ whnfIO (fun1 (\i -> return ()))
  , bench "2" $ whnfIO (fun2 (\i -> return ()))
  ]
```

### -O0
```
benchmarking 1
time                 14.95 μs   (14.88 μs .. 15.04 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 14.96 μs   (14.91 μs .. 15.04 μs)
std dev              210.7 ns   (150.5 ns .. 332.3 ns)
variance introduced by outliers: 10% (moderately inflated)

benchmarking 2
time                 12.20 μs   (12.18 μs .. 12.21 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 12.15 μs   (12.13 μs .. 12.17 μs)
std dev              62.67 ns   (54.52 ns .. 73.09 ns)
```
### -O
```
benchmarking 1
time                 11.84 μs   (11.83 μs .. 11.85 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 11.85 μs   (11.84 μs .. 11.86 μs)
std dev              45.18 ns   (35.97 ns .. 58.34 ns)

benchmarking 2
time                 13.23 μs   (13.21 μs .. 13.25 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 13.21 μs   (13.20 μs .. 13.23 μs)
std dev              52.67 ns   (40.70 ns .. 70.79 ns)
```

Data/Vector.hs
```diff
@@ -1063,6 +1064,18 @@ forM_ :: Monad m => Vector a -> (a -> m b) -> m ()
 {-# INLINE forM_ #-}
 forM_ = G.forM_
 
+-- | /O(n)/ Apply the monadic action to all elements and their indexes of the vector, yielding a
```
Bodigrim, line 1067: Other haddock comments and API (`findIndices`) in `vector` spell "indices" instead of "indexes" and AFAIK this is a preferred form in mathematical/technical context. If there is no particular reason to deviate, I suggest to stick to "indices".

Data/Vector.hs
```diff
@@ -1063,6 +1064,18 @@ forM_ :: Monad m => Vector a -> (a -> m b) -> m ()
...
+-- vector of results. Equivalent to @flip 'imapM'@.
```
Bodigrim, line 1068: This is nitpicking, but I suggest `@flip@ 'imapM'`or even better `'flip' 'imapM'`.

Data/Vector.hs
Bodigrim: Also, "their indexes of the vector" sounds a bit strange; "to all elements of the vector and their indices" is better.

fumieval: Fixed. I like "indices" better

Data/Vector.hs
fumieval: Fixed

lehins:
Looks great. All comments are addressed. @fumieval Thank you!

fumieval:
Thank you guys too for swift feedbacks!

PR #312, opened 2020-06-06
Add MonadFix instance for boxed vectors

### LINKED ISSUE #178: Add MonadFix instance for Vector

Shimuuar:
Supersedes #179. 
Fixes #178 

See both for earlier discussion. LGTM but I never really understood nor used MonadFix

@treeowl Do you still has interest in this PR? And by chance do you remember what does this passage means?

>  -- It's perfectly safe to use non-monadic indexing within
>  -- each element, as the result of indexing will be demanded
>  -- as soon as the vector is produced.


treeowl:
Why do we need a new PR? What was wrong with the previous one? I'll try to answer your questions today.

Shimuuar:
I resolved merge conflicts and I can't push to your repo.

P.S. If you want to amend anything in this PR ping me, I'll grant you push permissions in my fork

Shimuuar:
P.P.S reopened #179 just in case

Data/Vector.hs
```diff
@@ -381,6 +382,24 @@ instance MonadZip Vector where
   {-# INLINE munzip #-}
   munzip = unzip
 
+instance MonadFix Vector where
+  -- We take care to dispose of v0 as soon as possible.
+  -- We also avoid setting up the result vector to refer to
+  -- itself. These measures should prevent memory leaks.
+  -- It's perfectly safe to use non-monadic indexing within
+  -- each element, as the result of indexing will be demanded
+  -- as soon as the vector is produced.
```
treeowl, line 391: What I mean about this is that the use of `(!)` in line 399 is okay: it can't cause a vector to be kept alive by a thunk. Why? Because we don't *create* that vector until we need the value of the result of that indexing operation. This is as opposed to the `headM` we need to get `h`, where we'd otherwise keep `v0` alive in the first element of the result vector until that first element is forced by someone else.

Shimuuar: Thanks! 

I'll integrate that explanation into comment tomorrow. Right now it's somewhat cryptic

Shimuuar:
Sorry for the confusion. I should've word PR message better and least mention merge. I assumed that you gave up on PR after all these years

Data/Vector.hs
```diff
@@ -382,6 +383,26 @@ instance MonadZip Vector where
   {-# INLINE munzip #-}
   munzip = unzip
 
+-- | @since 0.13.0.0
```
Bodigrim, line 386: It would be nice to comment on the semantics of `MonadFix Vector`. It is similar to `MonadFix []` instance, isn't it? If it is, then what's wrong with definining it as below?
```
mfix f = V.fromList (mfix (V.toList . f))
```
I mean, the proposed implementation already `cons`es elements one by one, which is probably not very efficient.

treeowl: The proposed implementation does *not* cons elements one by one. It only uses `cons` *once*, and that `cons` should fuse with the following `generate`. A comment would be in order. Yes, semantics are the same as the `[]` instance, or close enough.

treeowl: The listy implementation could go in the instance documentation if you like.

Bodigrim: Ah, sorry, my bad, you are right. 
Yes, adding the "listy" implementation in the instance documentation sounds good to me.

Shimuuar:
I've updated comments

Shimuuar:
Thanks you, everyone. PR is finally merged

PR #314, opened 2020-06-07
Ignore upper bounds on vector in dependencies

Boarders:
Any dependency in the benchmarks that uses vector and has an upper bound will not allow this to build on master so we should ignore those bounds.


Bodigrim:
Thanks!

Issue #297, opened 2020-02-03
modern vector support ranges


cartazio:
the reason why vector has maintained such a long support window for as long as it has is largely because we literally couldn't robustly run CI on new ghc versions. That part is fixed by some fantastic insights by @lehins 

Theres a limit to the CPP a project can support. (while travis CI helps a huge amount, **it does burn folks out**, we all miss Tibbe on containers* land). For small, low surface area projects, its pretty easy to support a huge range in perpetuity, eg https://hackage.haskell.org/package/monad-ste is one i wrote a while ago that i doubt will need an update in the next decade unless ghc changes in strange new ways :).  

also, we genuinely aren't in the know about what ghc version ranges are genuinely being used outside of CI contexts.

# So i did some research 
* for the most gnarly and old distro in serious used: aka for RHEL 7 / Centos 7, you can get ghc 8.6.5 via  https://copr.fedorainfracloud.org/coprs/petersen/ghc-8.6.5/
(@juhp is the hero for those folks in the enterprise ).   Plus i'm genuinely unaware of any haskell engineering teams in those environments that dont use newer ghc versions than the one shipped by default in the distro for active development 
(perhaps @simonmar or @bodigrim or @juhp know of any? whether in their own work or via conversations with others, whats the oldest ghc they know folks actually currently use to ship the end applications?)

## bsds'
* openbsd ports tree has had the ghc ports tree version be >= 8.4 since 2018 (and they dont support versions more than 12 months)

* free bsd ports has had >= 8.4 since June 2018, and all releases from before then are past their end of life threshold., though @raichoo  can perhaps  add some color
## debians
*looking at debian, the oldest long term support release thats still going to have distro support in 6 months is https://wiki.debian.org/LTS debian 9 (buster), which ships with ghc 8.0, though via http://downloads.haskell.org/debian/, 8.6 and 8.8 ghc versions are available. 

* In 6 months, the oldest support debian release will be debian 10 (squeeze) which ships with ghc 8.4 (though again, via the adapted @hvr ppas, newer and old ghc are available )

## buntos

* on ubuntu the oldest flavor that has any sort of support atm is TrustyTahr (though i could be reading this wrong). Though if i'm reading it correctly, https://ubuntu.com/about/release-cycle, then thats more of a commercial security updates offering.  

* Xenial seems like the oldest LTS that gets more than security updates, and thats defaulting to 7.10,, though again the @hvr ppa's provide every ghc version under the sun 7.0 - 8.10.
 
[...]

PR #316, opened 2020-06-13
Make primitive vectors nominally roled

### LINKED ISSUE #277: Make sure next major version bump includes the correct role annotations for each vector type
Storable Vectors need a nominal role, and thats likely true for everything but boxed vectors! (afaict)

Shimuuar:
Reasoning is identical to one for storable vectors. See discussion in #224, and
PR #235.

Fixes #277

### DIFF
--- a/Data/Vector/Primitive.hs
+++ b/Data/Vector/Primitive.hs
@@ -137,6 +137,9 @@ module Data.Vector.Primitive (
 
   -- ** Other vector types
   G.convert,
+#if __GLASGOW_HASKELL__ >= 708
+  unsafeCoerceVector,
+#endif
 
   -- ** Mutable vectors
   freeze, thaw, copy, unsafeFreeze, unsafeThaw, unsafeCopy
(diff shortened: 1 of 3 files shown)


Data/Vector/Primitive.hs
```diff
@@ -137,6 +137,9 @@ module Data.Vector.Primitive (
 
   -- ** Other vector types
   G.convert,
+#if __GLASGOW_HASKELL__ >= 708
+  unsafeCoerceVector,
```
Bodigrim, line 141: Having conditional API is likely to make lifes of downstream dependencies more difficult. It may be better to export `unsafeCoerceVector` unconditionally and define it for old GHCs without `Coercible` constraint.

Data/Vector/Primitive.hs
```diff
@@ -183,11 +186,22 @@ import Data.Traversable ( Traversable )
...
+import Data.Coerce
+import Unsafe.Coerce
 import qualified GHC.Exts as Exts
 #endif
 
 #if __GLASGOW_HASKELL__ >= 708
-type role Vector representational
+type role Vector nominal
+
+-- | /O(1)/ Unsafely coerce a mutable vector from one element type to another,
```
Bodigrim, line 197: "a mutable" -> "an immutable"

changelog.md
```diff
@@ -15,7 +15,10 @@
      `type role MVector nominal representational` (previously, both arguments
      were `phantom`).
    * The role signature for `Data.Vector.Primitive.Vector` is now
-     `type role Vector representational` (previously, it was `phantom`).
+     `type role Vector nominal` (previously, it was `phantom`).
+     for `Data.Vector.Primitive.Mutable.MVector` is now
+     `type role MVector nominal nominal` (previous, both arguments were
+     `phantom`).
```
Bodigrim, line 21: "for" -> "For" (or preferably put "The role signature for" before) 
"previous" -> "previously".

Data/Vector/Primitive.hs
```diff
@@ -183,11 +186,22 @@ import Data.Traversable ( Traversable )
...
+-- representationally equal type. The operation just changes the type of the
+-- underlying pointer and does not modify the elements.
+--
+-- While it guarantees that elements of vector are representationally
+-- equal it doesn't guarantee that 'Prim' instances are memory-compatible.
```
Bodigrim, line 202: This is a subtle subject, and an unprepared reader may be well confused how types could happen to be "representationally equal", but not "memory-compatible". Here is my take (still far from perfect):

"Coerce a vector to another vector, when types of underlying elements are 'Coercible' to each other (e. g., newtypes). This is unsafe, because 'Prim' instances of newtypes can potentially still have different memory representations."

Shimuuar:
@Bodigrim I updated documentation but left CPP as it is. I'd like to revisit #297 before doing anything about it.

PR #317, opened 2020-06-14
Drop support for GHC<=7.8

Shimuuar:
As result of discussion in #297


Shimuuar:
I've rebased PR on top of master. No one complained so I'm merging it

PR #319, opened 2020-06-20
Fix rewrite rule for zipWithM

### LINKED ISSUE #252: Possible bug?: zipWithM xs xs rule
Hello. I've been lately reading the source of this package, and I found the rules like
``` Haskell
{-# RULES

"zipWithM xs xs [Vector.Stream]" forall f xs.
  zipWithM f xs xs = mapM (\x -> f x x) xs   #-}
```
in Data.Vector.Fusion.Stream.Monadic and Data.Vector.Fusion.Bundle.Monadic. These rules look like bugs to me, since the meanings of the two sides of the equation are different. Note that streams/bundles in these modules are *monadic* --- the left hand side duplicates and interleaves the actions in the two copies of the stream xs, and the right hand side executes each of the actions only once.

One possible interpretation is that these Monadic Fusion modules are _internal_ and it is intended only for use from vector modules, but it does not seem to be documented. What do you think?

Shimuuar:
For description of bug see #252. Fix is rather simple RULE now only matches pure streams which is safe since there's no chance of performing some effect only once instead of twice.

In order to be able to do so `lift` was moved from Bundle to Bundle.Monadic

Fixes #252

### DIFF
--- a/Data/Vector/Fusion/Bundle.hs
+++ b/Data/Vector/Fusion/Bundle.hs
@@ -80,7 +80,7 @@ import Data.Vector.Generic.Base ( Vector )
 import Data.Vector.Fusion.Bundle.Size
 import Data.Vector.Fusion.Util
 import Data.Vector.Fusion.Stream.Monadic ( Stream(..), Step(..) )
-import Data.Vector.Fusion.Bundle.Monadic ( Chunk(..) )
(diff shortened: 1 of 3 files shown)


Bodigrim:
LGTM, except GHC 7.10 build.

PR #320, opened 2020-06-20
Reexport PrimMonad from mutable vectors

### LINKED ISSUE #262: Missing API master issue

Shimuuar:
This allows to avoid extra import from primitive (and adding it to cabal file)
It's also in line with reexposrt of Storable and Prim type classes

Fixes #262

### DIFF
--- a/Data/Vector/Generic/Mutable.hs
+++ b/Data/Vector/Generic/Mutable.hs
@@ -13,7 +13,7 @@
 
 module Data.Vector.Generic.Mutable (
   -- * Class of mutable vector types
-  MVector(..),
+  MVector(..), PrimMonad(..),
 
   -- * Accessors
 
--- a/Data/Vector/Generic/Mutable.hs
+++ b/Data/Vector/Generic/Mutable.hs
@@ -13,7 +13,7 @@
 
 module Data.Vector.Generic.Mutable (
   -- * Class of mutable vector types
-  MVector(..), PrimMonad(..),
+  MVector(..),
 
   -- * Accessors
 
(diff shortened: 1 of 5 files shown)


lehins:
@Shimuuar What do you think about also re-exporting `RealWorld`?

It is not too uncommon to see: `(MonadIO m, PrimMonad m, PrimState m ~ RealWorld) =>`

Shimuuar:
Why not? These modules imported qualified so name pollution is not a problem

Data/Vector/Generic/Mutable.hs
```diff
@@ -13,7 +13,7 @@
 
 module Data.Vector.Generic.Mutable (
   -- * Class of mutable vector types
-  MVector(..),
+  MVector(..), PrimMonad(..), RealWorld,
```
Bodigrim, line 16: It's a bit weird to export `PrimMonad(..)` / `RealWorld` under "Class of mutable vector types", and render almost at the very top of haddocks. What about moving them to a new section "Re-exported entities" at the bottom of export list?

Shimuuar: Good point. I moved both to bottom of export list

PR #322, opened 2020-06-20
Add specializations for all functions from Generic

Shimuuar:
Progress is tracked in #299

### DIFF
--- a/Data/Vector.hs
+++ b/Data/Vector.hs
@@ -153,6 +153,9 @@ module Data.Vector (
   scanr, scanr', scanr1, scanr1',
   iscanr, iscanr',
 
+  -- ** Comparisons
+  eqBy, cmpBy,
+
   -- * Conversions
 
   -- ** Lists
@@ -1748,6 +1751,22 @@ scanr1 = G.scanr1
 scanr1' :: (a -> a -> a) -> Vector a -> Vector a
 {-# INLINE scanr1' #-}
 scanr1' = G.scanr1'
+
+-- Comparisons
+-- ------------------------
+
+-- | /O(n)/ Check if two vectors are equal using supplied equality
+-- predicate.
+eqBy :: (a -> b -> Bool) -> Vector a -> Vector b -> Bool
+{-# INLINE eqBy #-}
+eqBy = G.eqBy
+
+-- | /O(n)/ Compare two vectors using supplied comparison function for
+-- vector elements. Comparison works same as for lists.
+--
+-- > cmpBy compare == compare
+cmpBy :: (a -> b -> Ordering) -> Vector a -> Vector b -> Ordering
+cmpBy = G.cmpBy
 
 -- Conversions - Lists
 -- ------------------------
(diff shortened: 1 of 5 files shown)


Shimuuar:
Now all differences between Generic and rest of immutable vectors are accounted for. There're still mutable vectors. But that's much bigger fish to fry.

Shimuuar:
> I don't feel like this is 100% correct wording for cmpBy functions:

Ah! That's case of "I copy-paste, therefore I don't think". Equality is remnant of `eqBy` haddock

Data/Vector.hs
```diff
@@ -1743,6 +1746,19 @@ scanr1' :: (a -> a -> a) -> Vector a -> Vector a
...
+-- Comparisons
+-- ------------------------
+
+-- | /O(n)/ Check if two vectors are equal using supplied equality
+-- predicate.
+eqBy :: (a -> a -> Bool) -> Vector a -> Vector a -> Bool
+{-# INLINE eqBy #-}
+eqBy = G.eqBy
+
+-- | /O(n)/ Check if two vectors are equal using supplied comparison function
```
Bodigrim, line 1758: It's important to mention here and above what happens for vectors of different length.

Shimuuar: I finally updated haddocks for cmpBy

Data/Vector.hs
```diff
@@ -1746,6 +1749,22 @@ scanr1' :: (a -> a -> a) -> Vector a -> Vector a
...
+-- predicate.
+eqBy :: (a -> a -> Bool) -> Vector a -> Vector a -> Bool
+{-# INLINE eqBy #-}
+eqBy = G.eqBy
+
+-- | /O(n)/ Compare two vectors using supplied comparison function for
+-- vector elements. Comparison works same as for lists.
+--
+-- > cmpBy compare == compare
+cmpBy :: (a -> a -> Ordering) -> Vector a -> Vector a -> Ordering
```
Bodigrim, line 1765: Shouldn't it be 
```haskell
cmpBy :: (a -> b -> Ordering) -> Vector a -> Vector b -> Ordering
```
?

lehins: Good catch. That is the type of `G.cmpBy`, so it seems like it should apply to specialized versions as well.

Shimuuar: I specialized these to `a -> a -> Ordering` because almost always vectors of the same type are compared and if someone needs most generic case there's Data.Vector.Generic.cmpBy

Bodigrim: I do not see a good reason to restrict specialized versions in such way. This is both less powerful and less expected.

Shimuuar: I thought that `a -> b -> Ordering` is somewhat unexpected. Anyway I changed specializations to `a -> b -> Ordering`

Bodigrim: Could you please do the same for `eqBy`?

Shimuuar: Done. Thank you for noticing things.

PR #323, opened 2020-06-27
Fix fusion of `<$` operator

lehins:
I've tried both of the suggested approaches from #248 
```haskell
  {-# INLINE (<$) #-}
  (<$) a v = replicate (length v) a
```

and this one https://github.com/haskell/vector/pull/248#issuecomment-583745720 together with
```haskell
  {-# INLINE (<$) #-}
  (<$) x = G.unstream . (x <$) . G.stream
```

None of them prevent this example from diverging:
```haskell
  let v = V.unfoldr (\x -> Just (x, x + 1)) (0 :: Int)
  print $ V.take 4 ((1 :: Int) <$ v) -- diverges
```

Default implementation for `<$` is identical to the one in this PR, but because it does not have an `INLINE` pragma it also fails to fuse.

Approach in this PR is the only one I found that produces the desired output:
```
[1,1,1,1]
```

In other words, this PR does not change the semantics, but fixes fusion for `<$` operator.

### DIFF
--- a/Data/Vector.hs
+++ b/Data/Vector.hs
@@ -340,6 +340,9 @@ instance Functor Vector where
   {-# INLINE fmap #-}
   fmap = map
 
+  {-# INLINE (<$) #-}
+  (<$) = map . const
+
 instance Monad Vector where
   {-# INLINE return #-}
   return = Applicative.pure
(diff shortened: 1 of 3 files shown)


lehins:
@Shimuuar Done. Added the same implementation for both `Bundle` and `Stream`

Shimuuar:
It looks ready for merge to me. I would wait a bit to let @Bodigrim take a look at it

PR #325, opened 2020-07-12
Switch benchmarks from criterion to gauge

### LINKED ISSUE #321: Switch benchmarks from criterion to gauge
At the moment benchmarks are pretty much useless, because `criterion` depends on `vector` not only directly, but also transitively on many occasions. 
```
In the dependencies for aeson-1.4.7.1:
    vector-0.13.0.1 from stack configuration does not match >=0.12.0.1 && <0.13  (latest matching
                    version is 0.12.1.2)
needed due to vector-benchmarks-0.10.10 -> aeson-1.4.7.1

In the dependencies for cassava-0.5.2.0(-bytestring--lt-0_10_4):
    vector-0.13.0.1 from stack configuration does not match >=0.8 && <0.13  (latest matching version
                    is 0.12.1.2)
needed due to vector-benchmarks-0.10.10 -> cassava-0.5.2.0

In the dependencies for microstache-1.0.1.1:
    vector-0.13.0.1 from stack configuration does not match >=0.11 && <0.13  (latest matching version
                    is 0.12.1.2)
needed due to vector-benchmarks-0.10.10 -> microstache-1.0.1.1

In the dependencies for vector-algorithms-0.8.0.3:
    vector-0.13.0.1 from stack configuration does not match >=0.6 && <0.13  (latest matching version
                    is 0.12.1.2)
needed due to vector-benchmarks-0.10.10 -> vector-algorithms-0.8.0.3

In the dependencies for vector-binary-instances-0.2.5.1:
    vector-0.13.0.1 from stack configuration does not match >=0.6 && <0.13  (latest matching version
                    is 0.12.1.2)
needed due to vector-benchmarks-0.10.10 -> vector-binary-instances-0.2.5.1

In the dependencies for vector-th-unbox-0.2.1.7:
    vector-0.13.0.1 from stack configuration does not match >=0.7.1 && <0.13  (latest matching
                    version is 0.12.1.2)
needed due to vector-benchmarks-0.10.10 -> vector-th-unbox-0.2.1.7
```
I suggest we switch from `criterion` to `gauge` to simplify this sort of things.

----

(In the long term one could probably patch `gauge` to use `primitive` instead of `vector`)

Bodigrim:
Closes #321 

I also switched generation of random vectors from `mwc-random` to `random` to avoid building `math-functions` and `mwc-random` (both of them depend on `vector`). Now it is only `gauge` itself which needs to be rebuilt.


Bodigrim:
I also forked `gauge` to use `primitive` instead of `vector` when statistical analysis is disabled, providing even faster (but less accurate) feedback loop. 

```yaml
resolver: nightly-2020-05-03
packages:
- .
- benchmarks
extra-deps:
- random-1.2.0
- splitmix-0.1
- github: Bodigrim/hs-gauge
  commit: bf685c5
flags:
  gauge:
    analysis: false
```

Shimuuar:
> when statistical analysis is disabled

criterion/gauge are not very accurate as it is. Right now I'm studying now precise error estimates are from criterion/gauge. So far it look like they're underestimated by order of magnitude

PR #327, opened 2020-08-02
make dropWhile fusion adaptive

gksato:
Now dropWhile's implementation uses stream
only in case the stream fusion is already running.
This is expected to reduce unnecessary re-allocation.
This is an improvement of #146, which does not admit stream fusion in any case.

I've executed the tests included in the package with GHC 8.6.5 and 8.8.3.
If you wish to have other tests done, please tell me.
 
cf. https://github.com/haskell/vector/issues/141
https://github.com/haskell/vector/issues/108

### DIFF
--- a/Data/Vector/Generic.hs
+++ b/Data/Vector/Generic.hs
@@ -1381,8 +1381,19 @@ takeWhile f = unstream . Bundle.takeWhile f . stream
 -- | /O(n)/ Drop the longest prefix of elements that satisfy the predicate
 -- without copying.
 dropWhile :: Vector v a => (a -> Bool) -> v a -> v a
-{-# INLINE dropWhile #-}
-dropWhile f = unstream . Bundle.dropWhile f . stream
+{-# INLINE_FUSED dropWhile #-}
+-- In the case that the argument is an actual vector,
+-- this is a faster solution than stream fusion.
+dropWhile f xs = case findIndex (not . f) xs of
+                   Just i  -> unsafeDrop i xs
+                   Nothing -> empty
+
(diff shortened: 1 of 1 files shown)


Shimuuar:
At a glance PR seems fine but I want to recheck that it works as advertised. I'm always become very cautious when RULES are involved.

gksato:
Agreed. Please tell me if I can help you in any way with rechecking.

Data/Vector/Generic.hs
```diff
@@ -1381,8 +1381,19 @@ takeWhile f = unstream . Bundle.takeWhile f . stream
...
-dropWhile f = unstream . Bundle.dropWhile f . stream
+{-# INLINE_FUSED dropWhile #-}
+-- In the case that the argument is an actual vector,
+-- this is a faster solution than stream fusion.
+dropWhile f xs = case findIndex (not . f) xs of
+                   Just i  -> unsafeDrop i xs
+                   Nothing -> empty
+
+-- If the argument to 'dropWhile' comes from a stream,
+-- we need to avoid unnecessary allocation.
```
Shimuuar, line 1392: I think it's better to say that we want to avoid creating a new vector. Comment doesn't explain what is being allocated. 

Also please add mention that  `new (New.unstream ` is body of unstream.

gksato:
Alrighty. Adding a commit soon.

gksato:
Now I've committed and pushed a change to the comment. If you'd like a rebase, squash, etc, please tell me.

Shimuuar:
Thanks! I'll just squash merged PR.

PR #231, opened 2018-12-12
Improve "uninitialised element" error message

alexwl:
Compaction ([http://hackage.haskell.org/package/ghc-compact-0.1.0.0/docs/GHC-Compact.html#v:compact](http://hackage.haskell.org/package/ghc-compact-0.1.0.0/docs/GHC-Compact.html#v:compact)) of an immutable vector may crash with the confusing "uninitialised element" error message [https://github.com/haskell/vector/issues/220](https://github.com/haskell/vector/issues/220):

```haskell
module Main where

import qualified Data.Vector as V
import GHC.Compact

main :: IO ()
main = do
  c <- compact $ V.fromList [1, 2, 3]
  pure ()
```

```
*** Exception: Data.Vector.Mutable: uninitialised element
```

As far as I understand, there is no simple way to fix this. However, there is an easy workaround: use the `force` function.

This code doesn't crash:

```haskell
c <- compact $ V.force $ V.fromList [1, 2, 3]
```

The PR adds a suggestion to use the `force` function in the error message.

PR #328, opened 2020-08-11
Disambiguate where the suggested 'force' function comes from

nh2:
Follow-up to PR #231.

Users could easily be misled that it's `Control.DeepSeq.force`, not `Data.Vector.force`,
and the former does not fix the issue at all.

CC @recursion-ninja @enobayram

### DIFF
--- a/Data/Vector/Mutable.hs
+++ b/Data/Vector/Mutable.hs
@@ -193,7 +193,7 @@ loopM !n k = let
   in go 0
 
 uninitialised :: a
-uninitialised = error "Data.Vector.Mutable: uninitialised element. If you are trying to compact a vector, use the 'force' function to remove uninitialised elements from the underlying array."
+uninitialised = error "Data.Vector.Mutable: uninitialised element. If you are trying to compact a vector, use the 'Data.Vector.force' function to remove uninitialised elements from the underlying array."
 
 -- Length information
 -- ------------------


lehins:
@nh2 Good catch. However this error message is only applicable to boxed vectors, so maybe clarifying with: 

`... use the 'Data.Vector.force' function ...`

 instead of 

`... use the 'force' function (from your vector's module) ...`

nh2:
@lehins Ok, updated!

recursion-ninja:
Are you kidding me? I  ***𝚆𝙰𝚂𝙽'𝚃***  supposed to use `Control.DeepSeq.force`? For 2 years I've been repeatedly bitten by this class of issues from the `vector` package. This is a much needed change to the error message! Thanks for making the clarification "hard-coded" into the error message. Now other's should have to struggle so much.

lehins:
@recursion-ninja I know it's been two years, but to be fair you did make a https://github.com/haskell/vector/pull/231#issuecomment-509736562 on the PR that explicitly mentions `V.force`:

> This code doesn't crash:
> `c <- compact $ V.force $ V.fromList [1, 2, 3] `

I also hope you meant to say: "Now other's should **not** have to struggle so much." :smile:

That said, the first time I saw that issue I also thought the error message referred to the `Control.DeepSeq.force`, so it is indeed a very nice change to the error message.

recursion-ninja:
I did mean *shouldn't* (edited). 

I recall stumbling back and forth between the two `force` methods, reverting to `Control.DeepSeq.force` during some refactorings.

PR #329, opened 2020-08-28
Add generic function `catMaybes`

carlosdagos:
Add `catMaybes` which provides same function as the one found in `Data.Maybe`.

I thought this would be a useful function to have. Unfortunately the tests give me a type error when I added the property (in all honesty I don't understand the `VectorContext` constraint right now). But I thought that I'd hear thoughts about adding this before dedicating more time to it. 

Thanks!

### DIFF
--- a/Data/Vector.hs
+++ b/Data/Vector.hs
@@ -115,7 +115,7 @@ module Data.Vector (
 
   -- ** Filtering
   filter, ifilter, uniq,
-  mapMaybe, imapMaybe,
+  mapMaybe, imapMaybe, catMaybes,
   filterM,
   takeWhile, dropWhile,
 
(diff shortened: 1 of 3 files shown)


lehins:
The only module where `catMaybes` is valid is `Data.Vector`, so feel free to remove it from other modules, including `Data.Vector.Generic`. Reason for this is very simple, there are no instances of `Prim`, `Unbox` or `Storable` for `Maybe a`

carlosdagos:
Done @lehins. Thanks. I actually left the modification to =`Data.Vector.Fusion.Stream.Monadic` as I think it could be useful there. But I am happy to amend that if need be as well. Cheers.

PR #333, opened 2020-10-10
Treeowl map maybe m

### LINKED ISSUE #183: mapMaybeM

Shimuuar:
This is #226 rebased on top of master and with haddocks. Only think holding #226 is lack of documentation so I think now it's ready to merge. 

Fixes #183

### DIFF
--- a/Data/Vector/Generic.hs
+++ b/Data/Vector/Generic.hs
@@ -97,9 +97,9 @@ module Data.Vector.Generic (
   -- * Working with predicates
 
   -- ** Filtering
-  filter, ifilter, uniq,
+  filter, ifilter, filterM, uniq,
   mapMaybe, imapMaybe,
-  filterM, mapMaybeM, imapMaybeM,
+  mapMaybeM, imapMaybeM,
   takeWhile, dropWhile,
 
   -- ** Partitioning
--- a/Data/Vector/Generic.hs
+++ b/Data/Vector/Generic.hs
@@ -1382,6 +1382,7 @@ mapMaybeM f = unstreamM . Bundle.mapMaybeM f . stream
 -- Discards elements returning Nothing.
 imapMaybeM :: (Monad m, Vector v a, Vector v b)
       => (Int -> a -> m (Maybe b)) -> v a -> m (v b)
+{-# INLINE imapMaybeM #-}
 imapMaybeM f = unstreamM . Bundle.mapMaybeM (\(i, a) -> f i a) . Bundle.indexed . stream
 
 -- | /O(n)/ Yield the longest prefix of elements satisfying the predicate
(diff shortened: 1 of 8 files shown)


Data/Vector/Generic.hs
```diff
@@ -1372,6 +1372,18 @@ filterM :: (Monad m, Vector v a) => (a -> m Bool) -> v a -> m (v a)
...
+-- discard elements returning Nothing.
+mapMaybeM :: (Monad m, Vector v a, Vector v b) => (a -> m (Maybe b)) -> v a -> m (v b)
+{-# INLINE mapMaybeM #-}
+mapMaybeM f = unstreamM . Bundle.mapMaybeM f . stream
+
+-- | /O(n)/ Apply monadic function to each element of vector and its index.
+-- Discards elements returning Nothing.
+imapMaybeM :: (Monad m, Vector v a, Vector v b)
+      => (Int -> a -> m (Maybe b)) -> v a -> m (v b)
+imapMaybeM f = unstreamM . Bundle.mapMaybeM (\(i, a) -> f i a) . Bundle.indexed . stream
```
lehins, line 1385: Missing the inline pragma:
```suggestion
{-# INLINE imapMaybeM #-}
imapMaybeM f = unstreamM . Bundle.mapMaybeM (\(i, a) -> f i a) . Bundle.indexed . stream
```

Data/Vector/Generic.hs
```diff
@@ -99,7 +99,7 @@ module Data.Vector.Generic (
   -- ** Filtering
   filter, ifilter, uniq,
   mapMaybe, imapMaybe,
-  filterM,
+  filterM, mapMaybeM, imapMaybeM,
```
lehins, line 102: Not terribly important, but for consistency sake. All other modules export `mapMaybeM` and `imapMaybeM` before `filterM` and after `mapMaybeM`, `imapMaybe`.

```suggestion
  mapMaybeM, imapMaybeM,
  filterM,
```

lehins (approved):
Couple of minor comments, but otherwise looks great.

Data/Vector/Generic.hs
Shimuuar: Data.Vector wasn't consistent as well. I arranged everything in the same way. I also put `filterM` after `filter` & `ifilter`. I think it makes sense to keep all filter variants together.

Bodigrim (approved):

Issue #315, opened 2020-06-09
Deriving of unboxed vector for newtypes


Shimuuar:
Problem is very simple. Newtype deriving doesn't work since roles were introduced to GHC.

```haskell
{-# LANGUAGE DerivingStrategies         #-}
{-# LANGUAGE GeneralizedNewtypeDeriving #-}
{-# LANGUAGE MultiParamTypeClasses      #-}
{-# LANGUAGE StandaloneDeriving         #-}
{-# LANGUAGE TypeFamilies               #-}
import qualified Data.Vector.Unboxed         as U
import qualified Data.Vector.Unboxed.Mutable as MU
import qualified Data.Vector.Generic         as G
import qualified Data.Vector.Generic.Mutable as GM
newtype X = X Int

newtype instance U.Vector     X = V_X (U.Vector Int)
newtype instance MU.MVector s X = M_X (MU.MVector s Int)

deriving newtype instance G.Vector   U.Vector   X
deriving newtype instance GM.MVector MU.MVector X 
```

 So this simple program fails with:

```
    • Couldn't match representation of type ‘m1 (MU.MVector
                                                   (PrimState m1) Int)’
                               with that of ‘m1 (MU.MVector
                                                   (PrimState m1) X)’
        arising from a use of ‘GHC.Prim.coerce’
    ...
```

Reason for that are very obvious. Methods of type class has type `∀m. PrimMonad m => ... -> m a` and GHC can't coerce `m Int` to `m X` because it can't assume that Int has representational/phantom role. 

This problem could be fixed by changing `∀m. PrimMonad ⇒ m` type parameter to `∀s. ST s`. Since `m` only appears in positive positions both approaches are equivalent and every instance that could be written in one style could be written in another.

This is of course breaking change but breakage should be relatively limited. Type class methods are not meant to be used directly and custom instances should continue to compile

Bodigrim:
What would happen to mutable vectors, manipulated in `IO` monad, under this proposal?

Shimuuar:
Do you mean ones that use IO in implementation (Storable)? I think nothing. They need to use unsafePrimToPrim anyway. It will work just fine for converting to ST.

User facing API shouldn't change at all. D.V.Generic.Mutable should continue to use PrimMonad

Shimuuar:
Yes. Exactly this.

lehins:
This looks legit to me.

Bodigrim:
And what would be the definition of `clear`?

[...]

PR #335, opened 2020-10-19
Use ST instead of polymorphic m in Vector/MVector type classes

Shimuuar:
This is implementation of idea from #315. Instead of using polymorphic `m` in methods of MVector/Vector type class use ST and lift it to PrimMonad at use sites. It's probably easier to illustrate idea with few diffs:

```
-  basicUnsafeNew   :: PrimMonad m => Int -> m (v (PrimState m) a)
+  basicUnsafeNew   :: Int -> ST s (v s a)
```
```
 unsafeRead v i = UNSAFE_CHECK(checkIndex) "unsafeRead" i (length v)
+               $ stToPrim
                $ basicUnsafeRead v i
```

Main motivation for change is to allow GND for those type classes. It's partial success. In following code snippet:

```haskell
newtype instance MVector s (Sum a) = MV_Sum (MVector s a)
newtype instance Vector    (Sum a) = V_Sum  (Vector a)
deriving instance Unbox a => G.Vector  Vector  (Sum a)
deriving instance Unbox a => M.MVector MVector (Sum a)
```

MVector instance derived successfully. Vector fails with:
```
    • Couldn't match representation of type ‘m1 a’
                               with that of ‘m1 (Sum a)’
```
because of `basicUnsafeIndexM  :: Monad m => v a -> Int -> m a`. GHC cannot coerce between `m a` and `m (Sum a)`. I'm a bit at loss here.

### DIFF
--- a/Data/Vector/Generic/Base.hs
+++ b/Data/Vector/Generic/Base.hs
@@ -25,6 +25,7 @@ module Data.Vector.Generic.Base (
 import           Data.Vector.Generic.Mutable.Base ( MVector )
 import qualified Data.Vector.Generic.Mutable.Base as M
 
+import Control.Monad.ST
 import Control.Monad.Primitive
 
 -- | @Mutable v s a@ is the mutable version of the pure vector type @v a@ with
@@ -59,13 +60,13 @@ class MVector (Mutable v) a => Vector v a where
   -- Unsafely convert a mutable vector to its immutable version
   -- without copying. The mutable vector may not be used after
   -- this operation.
-  basicUnsafeFreeze :: PrimMonad m => Mutable v (PrimState m) a -> m (v a)
+  basicUnsafeFreeze :: Mutable v s a -> ST s (v a)
 
   -- | /Assumed complexity: O(1)/
   --
   -- Unsafely convert an immutable vector to its mutable version without
   -- copying. The immutable vector may not be used after this operation.
-  basicUnsafeThaw :: PrimMonad m => v a -> m (Mutable v (PrimState m) a)
+  basicUnsafeThaw :: v a -> ST s (Mutable v s a)
 
   -- | /Assumed complexity: O(1)/
   --
@@ -117,7 +118,7 @@ class MVector (Mutable v) a => Vector v a where
   --
   -- Default definition: copying basic on 'basicUnsafeIndexM' and
   -- 'basicUnsafeWrite'.
-  basicUnsafeCopy :: PrimMonad m => Mutable v (PrimState m) a -> v a -> m ()
+  basicUnsafeCopy :: Mutable v s a -> v a -> ST s ()
 
   {-# INLINE basicUnsafeCopy #-}
   basicUnsafeCopy !dst !src = do_copy 0
--- a/Data/Vector/Generic/Base.hs
+++ b/Data/Vector/Generic/Base.hs
@@ -24,6 +24,7 @@ module Data.Vector.Generic.Base (
 
 import           Data.Vector.Generic.Mutable.Base ( MVector )
 import qualified Data.Vector.Generic.Mutable.Base as M
+import           Data.Vector.Fusion.Util (Box(..), liftBox)
 
 import Control.Monad.ST
 import Control.Monad.Primitive
@@ -106,7 +107,7 @@ class MVector (Mutable v) a => Vector v a where
   -- which does not have this problem because indexing (but not the returned
   -- element!) is evaluated immediately.
   --
-  basicUnsafeIndexM  :: Monad m => v a -> Int -> m a
+  basicUnsafeIndexM  :: v a -> Int -> Box a
 
   -- |  /Assumed complexity: O(n)/
   --
@@ -126,7 +127,7 @@ class MVector (Mutable v) a => Vector v a where
       !n = basicLength src
 
       do_copy i | i < n = do
-                            x <- basicUnsafeIndexM src i
+                            x <- liftBox $ basicUnsafeIndexM src i
                             M.basicUnsafeWrite dst i x
                             do_copy (i+1)
                 | otherwise = return ()
--- a/Data/Vector/Unboxed/Base.hs
+++ b/Data/Vector/Unboxed/Base.hs
@@ -16,7 +16,7 @@
 --
 
 module Data.Vector.Unboxed.Base (
-  MVector(..), IOVector, STVector, Vector(..), Unbox
+  MVector(..), IOVector, STVector, Vector(..), Unbox, UnboxViaPrim(..)
 ) where
 
 import qualified Data.Vector.Generic         as G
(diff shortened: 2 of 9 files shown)


Shimuuar:
I said that I don't know what to do and then I used same trick: use box monad in order to express strictness in indexing and keep indexed value unevaluated. Everything works except GHC8.0 can't derive instances for some reason. While this Box trick seems to work I didn't test whether it really works.

I also want to test whether this approach works with DerivingVia. GND is surely nice but I'm not convinced that it alone make this change valuable enough.

Shimuuar:
In the end idea of using concrete monads in the definition of vector's type classes worked out nicely. `basicWhatever` methods which used `PrimMonad m` now are defined in terms of `ST`, and basicIndexM is defined in terms of `Box`. Rest of API is not changed at all. 

1. GND works. It's now possible to define unbox instances for all newtypes using it. (Except GHC8.0 fails with weird error, and we'll have to wait until we drop support for it)

2. DerivingVia works too. As an example:
```haskell
newtype instance MVector s Int = MV_Int (P.MVector s Int)
newtype instance Vector    Int = V_Int  (P.Vector    Int)
instance Unbox Int
deriving via (UnboxViaPrim Int) instance M.MVector MVector Int
deriving via (UnboxViaPrim Int) instance G.Vector Vector Int
```
3, `Box` tricked for indexM works too:
```
*Data.Vector> _ <- indexM (Data.Vector.fromList [undefined]) 0
*Data.Vector> _ <- indexM undefined 0
*** Exception: Prelude.undefined
CallStack (from HasCallStack):
  error, called at libraries/base/GHC/Err.hs:80:14 in base:GHC.Err
  undefined, called at <interactive>:5:13 in interactive:Ghci1
```

4. As small test for breakage I built statistics with this change. It and its dependencies has both TH and handwritten instances. All builds without any problems

Overall I think this is big improvement for vector usability.

lehins:
@Shimuuar This is great news! There is only one concern I have about GND and unboxing is that deriving does not add INLINE pragmas and I have seen this happen before where GHD causes derived class functions from not being inlined. (Here is a real life example where I saw this being a problem: commercialhaskell/rio#166) If the same happens with deriving unboxed vector it can have some serious performance implications. I think it would be great to have some benchmarks comparison of at least few functions of derived vs underived approach. Have you tried checking performance or differences in compiled core?

Other than the above concern I think this is a great improvement. I'll do a proper PR review either today or the latest tomorrow, just in case if there is anything I can spot.

Shimuuar:
Good point! I would've expect that GHC inlines coerced methods eagerly. I'll look into it on the weeekends

Shimuuar:
I've converted all existing benchmarks which use `Double` to `CDouble`. There's no measurable difference:

#### Double:
```
quickhull                                mean 85.75 ms  ( +- 994.5 μs  )
spectral                                 mean 3.183 ms  ( +- 35.50 μs  )
tridiag                                  mean 99.24 ms  ( +- 1.458 ms  )
findIndexR                               mean 189.7 μs  ( +- 2.424 μs  )
findIndexR_naïve                         mean 109.1 ms  ( +- 3.609 ms  )
findIndexR_manual                        mean 352.2 μs  ( +- 10.20 μs  )
```

#### CDouble:
```
quickhull                                mean 85.88 ms  ( +- 967.8 μs  )
spectral                                 mean 3.180 ms  ( +- 92.69 μs  )
tridiag                                  mean 99.59 ms  ( +- 2.563 ms  )
findIndexR                               mean 194.0 μs  ( +- 7.610 μs  )
findIndexR_naïve                         mean 109.8 ms  ( +- 4.609 ms  )
findIndexR_manual                        mean 349.3 μs  ( +- 5.304 μs  )
```

P.S. We don't have instances for types from Foreight.C.Types. Looks like oversight for me

Data/Vector/Generic/Base.hs
```diff
@@ -1,5 +1,5 @@
 {-# LANGUAGE Rank2Types, MultiParamTypeClasses, FlexibleContexts,
-             TypeFamilies, ScopedTypeVariables, BangPatterns #-}
+             TypeFamilies, ScopedTypeVariables, BangPatterns, DeriveFunctor #-}
```
lehins, line 2: I couldn't find where `DeriveFunctor` is being used in this PR? Also, according to ghc manual `DeriveFunctor` is available since ghc-7.10, if that extension is necessary for something are we dropping support for older ghc? If, yes, I am in favor, but we'll have to coordinate all the cleanup in a separate ticket.

lehins (approved):
@Shimuuar Besides few minor suggestions this PR looks great. Do you mind adding an entry to  change log that mentions this breaking change

lehins:
With regards to failing CI. I've seen doctests consistently fail on other projects when executed with ghc-8.10. I have not investigated the issue what is the cause of it, since doctests are mainly intended to check that examples are still correct. So we might wanna turn off doctests for ghc-8.10 until this issue resolves itself with future ghc+doctests.

Data/Vector/Unboxed/Base.hs
```diff
@@ -165,6 +165,90 @@ instance G.Vector Vector () where
...
+--
+-- Second example is essentially same but with a twist. Instead of
+-- using @Prim@ instance of data type, we use underlying instance of @Int@:
+--
+-- >>> import qualified Data.Vector.Unboxed         as U
+-- >>> import qualified Data.Vector.Primitive       as P
+-- >>> import qualified Data.Vector.Generic         as G
+-- >>> import qualified Data.Vector.Generic.Mutable as M
+-- >>>
+-- >>> newtype Bar = Bar Int deriving P.Prim
```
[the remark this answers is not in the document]
Shimuuar: > copy and paste error

It is, it is. :)

Data/Vector/Unboxed/Base.hs
```diff
@@ -560,3 +644,15 @@ deriveNewtypeInstances(Unbox (f (g a)), Compose f g a, f (g a), Compose, V_Compo
 
 #define DEFINE_INSTANCES
 #include "unbox-tuple-instances"
+
+
+-- $setup
+--
+-- >>> :set -XTypeFamilies
```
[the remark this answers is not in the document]
Shimuuar: Half of extensions are needed to derive Prim which is normally not done. I wonder if only extensions for deriving via should be listed in example text?..

Data/Vector/Unboxed/Base.hs
```diff
@@ -165,6 +165,90 @@ instance G.Vector Vector () where
 -- Primitive types
 -- ---------------
 
+-- | Newtype wrapper which allows to derive unboxed vector in term of
```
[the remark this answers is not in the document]
Shimuuar: Adding GND example is very good idea but it's probably better to add it in documentation o Unboxed module. Add section about deriving of unboxed vector.

Data/Vector/Generic/Base.hs
Shimuuar: That's remnant of `Box` monad, then I found that vector already has one and DeriveFunctor remained. I'll remove it.

Bodigrim:
> So we might wanna turn off doctests for ghc-8.10 until this issue resolves itself with future ghc+doctests.

Let's disable them for now, we've seen the same issue with `random` some time ago.

Data/Vector/Unboxed/Base.hs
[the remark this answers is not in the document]
Bodigrim: You can enable all extensions in one line. 
```
:set -XTypeFamilies -XDerivingVia ...
```
Let's list all extensions, it does not harm.

Data/Vector/Generic/Base.hs
Bodigrim: (We've already dropped GHC < 7.10)

lehins: @Bodigrim That's right, I remember now reviewing the PR #317 
Thanks for reminding me ;)

Shimuuar:
I've fixed everything except documentation on using GND for unboxed vectors.Will write it tomorrow

changelog.md
```diff
@@ -1,5 +1,10 @@
 # Changes in NEXT_VERSION
 
+ * Methods of type classes 'Data.Vector.Generic.Mutable.MVector' and
+   'Data.Vector.Generic.Vector' use concrete monads (ST,etc) being
+   polymorphic. This allows use of GND and deriving via to define
+   instances. Rest of API is unchanged and most existing instances should
+   compiler fine with new definitions.
```
lehins, line 7:
```suggestion
   compile fine with new definitions.
```

Shimuuar: Thanks!

Shimuuar:
I've added note about GND. I didn't change existing documentation.But it seems to be somewhat misleading:

> Implementing unboxed vectors for new data types can be very easy.

In particular this line seems to be out of place. BTW approach outlined there (and it's one used by vector-th-unbox) could be turned into deriving via with approach a la https://hackage.haskell.org/package/iso-deriving It's probably should be done in separate PR

lehins:
@Shimuuar Do you mind rebasing on master and squashing all the commits? Other than that I think we can merge it in

changelog.md
```diff
@@ -1,5 +1,10 @@
 # Changes in NEXT_VERSION
 
+ * Methods of type classes 'Data.Vector.Generic.Mutable.MVector' and
+   'Data.Vector.Generic.Vector' use concrete monads (ST,etc) being
```
Bodigrim, line 4: This is markdown, so backticks instead of single quotes please.

Shimuuar:
@lehins done

@Bodigrim thanks! I've updated changelog

Bodigrim:
Awesome stuff!

PR #336, opened 2020-10-21
Add equality that uses pointer equality as optimization

### LINKED ISSUE #239: Optimize Eq Data.Vector.Storable.Vector with a reference equality check?
We know that two vectors will be equal if they point to exactly the same range of the same `ForeignPtr`. Perhaps the `Eq` instance could exploit this to give faster termination for the case of `v == v` ?

Shimuuar:
Fixes #239 

Also adds doctests. Those don't test anything beyond fastEq but adding more should be easy

### DIFF
--- a/Data/Vector/Storable.hs
+++ b/Data/Vector/Storable.hs
@@ -129,6 +129,10 @@ module Data.Vector.Storable (
   -- ** Comparisons
   eqBy, cmpBy,
 
+  -- * Utilities
+  -- ** Comparisons
+  isSameVector,
+
   -- * Conversions
 
   -- ** Lists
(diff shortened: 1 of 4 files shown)


Data/Vector/Storable.hs
```diff
@@ -980,6 +984,20 @@ izipWith6 :: (Storable a, Storable b, Storable c, Storable d, Storable e,
...
+--
+-- >>> let xs = fromList [0/0::Double] in fastEq xs xs
+-- True
+--
+-- >>> let xs = fromList [0/0::Double] in xs == xs
+-- Falsec
+fastEq :: (Storable a, Eq a) => Vector a -> Vector a -> Bool
+fastEq v1@(Vector n1 ptr1) v2@(Vector n2 ptr2)
+  | n1 == n2 && ptr1 == ptr2 = True
+  | otherwise                = v1 == v2
```
Bodigrim, line 999: Could we rather expose 
```haskell
isSamePtr :: Vector a -> Vector a -> Bool
isSamePtr (Vector n1 ptr1) (Vector n2 ptr2) = n1 == n2 && ptr1 == ptr2
```
and let power users construct `fastEq` themselves?

This routine is faster only in a very specific case and is (marginally) slower otherwise. I'm not convinced this warrants `fastEq` name.

Shimuuar: That's good idea. It's probably better to call  it `isSameBuffer` since pointers could be same but length differ.

Shimuuar:
I seems that doctests work for GHC>=8.6. Older choke on `{-# UNPACK #-}` pragma. After this PR is merged adding doctest will be as simple as adding them

Data/Vector/Storable.hs
lehins: I'd rather call it `isSameVector`. In order to know that it is a pointer or a buffer you need to understand the internals, but `isSameVector` will be clear from the name what  it means. More over different offsets do means different vectors, but do not mean that the buffer or the pointer are different.

Shimuuar:
I've changed new function to `isSameVector`it's best name I think. Also I updated changelog including {i,}mapMaybeM.

Bodigrim (approved):

PR #337, opened 2020-10-27
Add generic foldMap

### LINKED ISSUE #263: No foldMap in Vector.Unboxed or Vector.Primitive
Both `Data.Vector.Primitive` and `Data.Vector.Unboxed` do not have a foldMap function. There is no particularly compelling technical reason this doesn't exist. It has the following implementation for example:
```haskell
foldMap :: forall m a . (Unbox a, Monoid m) => (a -> m) -> Vector a -> m
{-# INLINE foldMap #-}
foldMap f = foldr (mappend . f) mempty
```
```haskell
foldMap :: forall m a. (Prim a, Monoid m) => (a -> m) -> Vector a -> m
{-# INLINE foldMap #-}
foldMap f = foldr (mappend . f) mempty
```
Note that this is the default implementation in the Foldable typeclass. The main downside to adding this definition is that is adds a potential namespace clash but both of these modules are already drowning in such namespace clashes and so I don't see this as a serious objection.

Shimuuar:
Fixes #263. `foldMap` is implemented in terms of `foldr` and `foldMap'` in terms of `foldl'`

This is draft since I want to make sure that we agree on implementation and documentation and only then do copy-paste work. I certainly don't like haddocks as wrote them. Those are riddles fit for sphinx. `foldMap` is useful when one do need laziness but I can't put in words. Suggestions?

### DIFF
--- a/Data/Vector.hs
+++ b/Data/Vector.hs
@@ -129,6 +129,7 @@ module Data.Vector (
   -- * Folding
   foldl, foldl1, foldl', foldl1', foldr, foldr1, foldr', foldr1',
   ifoldl, ifoldl', ifoldr, ifoldr',
+  foldMap, foldMap',
 
   -- ** Specialised folds
   all, any, and, or,
@@ -200,7 +201,7 @@ import Prelude hiding ( length, null,
                         zipWith, zipWith3, zip, zip3, unzip, unzip3,
                         filter, takeWhile, dropWhile, span, break,
                         elem, notElem,
-                        foldl, foldl1, foldr, foldr1,
+                        foldl, foldl1, foldr, foldr1, foldMap,
                         all, any, and, or, sum, product, minimum, maximum,
                         scanl, scanl1, scanr, scanr1,
                         enumFromTo, enumFromThenTo,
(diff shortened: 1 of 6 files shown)


lehins:
Looks good to me. I don't think we need to dwell too much on explaining these functions since this is the concept from `base`. Maybe linking to corresponding functions would be enough for the unaware, eg "Same as 'Data.Foldable.foldMap', but for vectors"

Shimuuar:
I'd like to keep this PR for a few days. Maybe I'll come with something better. It's well known concept but I think it worth repeating

Data/Vector.hs
```diff
@@ -1474,6 +1475,23 @@ ifoldr' :: (Int -> a -> b -> b) -> b -> Vector a -> b
 {-# INLINE ifoldr' #-}
 ifoldr' = G.ifoldr'
 
+-- | @since NEXT
+-- /O(n)/ Map each element of the structure to a monoid, and combine
+-- the results. This function is implemented in terms of 'foldr' in
+-- the same way as default implementation in 'Foldable' type class.
```
Bodigrim, line 1481: We could avoid refering to implementation details saying "Map each element of the structure to a monoid, and combine results {lazily,strictly}, similar to `Data.Foldable.{foldMap,foldMap'}`".

Shimuuar: At first I removed foldl/foldr mentions as implementation detial. But then decided to restore it. Whether function will fuse or not depends whether foldl/foldr is used. It's unfortunate that implementation details leak like that

Shimuuar:
I tried and failed to come up with anything better. So I think PR is ready for merge

Data/Vector.hs
```diff
@@ -1475,6 +1476,24 @@ ifoldr' :: (Int -> a -> b -> b) -> b -> Vector a -> b
...
+-- 'Foldable' type cless. Note it's implemented in terms of 'foldr'
+-- and won't fuse with functions that traverse vector from left to
+-- right ('map', 'generate', etc.).
+foldMap :: (Monoid m) => (a -> m) -> Vector a -> m
+{-# INLINE foldMap #-}
+foldMap = G.foldMap
+
+-- | /O(n)/ 'foldMap' which is strict in accumulator. It uses same
+-- implementation as corresponding method of 'Foldable' type class.
+-- Note it's implemented in terms of 'foldl'' so it fuse in most
```
Bodigrim, line 1490:
```suggestion
-- Note it's implemented in terms of 'foldl'' so it fuses in most
```

Data/Vector/Generic.hs
```diff
@@ -1630,6 +1631,25 @@ ifoldr' :: Vector v a => (Int -> a -> b -> b) -> b -> v a -> b
...
+-- | /O(n)/ 'foldMap' which is strict in accumulator. It uses same
+-- implementation as corresponding method of 'Foldable' type class.
+-- Note it's implemented in terms of 'foldl'' so it fuse in most
+-- contexts.
+foldMap' :: (Monoid m, Vector v a) => (a -> m) -> v a -> m
+{-# INLINE foldMap' #-}
+foldMap' f = foldl' (\acc a -> acc `mappend` f a) mempty
+
+
+
```
Bodigrim, line 1652: This is a very minor nitpicking, but you put three empty lines here, two empty lines in `Data.Vector` and one line elsewhere.

Shimuuar:
@Bodigrim fixed everything you mentioned

Bodigrim (approved):

Issue #334, opened 2020-10-18
Enrich mutable API for vectors


Shimuuar:
There were quite a few proposal to enrich API of mutable vectors: #326, #203, #64, #175. Tris issue tries to collect all proposals from them in one nice list (and adds few on top of them)

### Folds & lookup

- [x] `foldlM :: (b -> a -> m b) -> b -> v s a -> m b`
- [x] `foldrM :: (a -> b -> m b) -> b -> v s a -> m b`
- [x] `mapM_ :: (a -> m ()) -> v s a -> m ()` / `forM_`
- [x] `imapM_ :: (Int -> a -> m ()) -> v s a -> m ()` / `forM_`
- [x] `ifoldlM` / `ifoldrM`
- [ ] `elemIndex`

I'm not sure that `foldlM_` & friends are worth adding. It's easy to ignore return result of fold anyway: `void`, `_ <- ...`. It however could be useful to add pure variants:

- [x] `foldl :: (b -> a -> b) -> b -> v s a -> m b`
- [x] `foldr :: (a -> b -> b) -> b -> v s a -> m b`

They come with performance bonus. We can perform entire fold in the ST monad and only lift to `m` at last moment. This way GHC only have to use bind from ST monad and it's good at optimizing it

### In place mutation

Naming could be slightly confusing here. Should in place map be called `mapM` or `mapM_`. Underscore usually means throwing away results of function and `mapM` builds new structure. I'm slightly lean towards `mapM` since it's in-place map

- [ ] `mapM :: (a -> m a) -> v s a -> m ()` / `forM`
- [ ] `map :: (a -> a) -> v s a -> m ()`  / for
- [ ] `imap :: (Int -> a -> a) -> v s a -> m ()`  / ifor
- [ ] `imapM :: (Int -> a -> m a) -> v s a -> m ()` / `iforM`
- [x] `modifyM :: v s a -> (a -> m a) -> Int -> m ()`
- [x] `previousPermutation :: (PrimMonad m, Ord e, MVector v e) => v (PrimState m) e -> m Bool` (to complement nextPermutation)
- [x] in-place `reverse`

### Creation

- [x] `generate :: Int -> (Int -> a) -> m (v s a)`
- [x] `generateM :: Int -> (Int -> m a) -> m (v s a)`

Comments? Objections? Proposals?

Bodigrim:
[...]

PR #338, opened 2020-10-27
Add generate, modifyM & folds for mutable vectors

Shimuuar:
See #334 and linked issues for discussion. Draft since functions were only added to generic module

- generate{,M}
- modifyM
- mapM_,imapM_, all variants of left folds 

These I think completely uncontroversial function

### DIFF
--- a/Data/Vector/Generic/Mutable.hs
+++ b/Data/Vector/Generic/Mutable.hs
@@ -639,7 +639,7 @@ generate n f = stToPrim $ generateM n (return . f)
 generateM :: (PrimMonad m, MVector v a) => Int -> (Int -> m a) -> m (v (PrimState m) a)
 {-# INLINE generateM #-}
 generateM n f
-  | n < 0     = new 0
+  | n <= 0    = new 0
   | otherwise = do
       vec <- new n
       let loop i | i >= n    = return vec
--- a/Data/Vector/Generic/Mutable.hs
+++ b/Data/Vector/Generic/Mutable.hs
@@ -638,12 +638,14 @@ generate n f = stToPrim $ generateM n (return . f)
 -- index. Iteration starts at index 0.
 generateM :: (PrimMonad m, MVector v a) => Int -> (Int -> m a) -> m (v (PrimState m) a)
 {-# INLINE generateM #-}
-generateM n f = do
-  vec <- new n
-  let loop i | i >= n    = return vec
-             | otherwise = do unsafeWrite vec i =<< f i
-                              loop (i + 1)
-  loop 0
+generateM n f
+  | n < 0     = new 0
+  | otherwise = do
+      vec <- new n
+      let loop i | i >= n    = return vec
+                 | otherwise = do unsafeWrite vec i =<< f i
+                                  loop (i + 1)
+      loop 0
 
 -- | Create a copy of a mutable vector.
 clone :: (PrimMonad m, MVector v a) => v (PrimState m) a -> m (v (PrimState m) a)
--- a/Data/Vector/Generic/Mutable.hs
+++ b/Data/Vector/Generic/Mutable.hs
@@ -934,7 +934,7 @@ ifoldr f b0 v = stToPrim $ ifoldrM (\i a b -> return $ f i a b) b0 v
 -- to each element and its index).
 ifoldr' :: (PrimMonad m, MVector v a) => (Int -> a -> b -> b) -> b -> v (PrimState m) a -> m b
 {-# INLINE ifoldr' #-}
-ifoldr' f b0 v = stToPrim $ ifoldrM (\i a b -> return $ f i a b) b0 v
+ifoldr' f b0 v = stToPrim $ ifoldrM' (\i a b -> return $ f i a b) b0 v
 
 -- | /O(n)/ Monadic fold.
 foldM :: (PrimMonad m, MVector v a) => (b -> a -> m b) -> b -> v (PrimState m) a -> m b
--- a/Data/Vector/Generic/Mutable.hs
+++ b/Data/Vector/Generic/Mutable.hs
@@ -30,7 +30,7 @@ module Data.Vector.Generic.Mutable (
   -- * Construction
 
   -- ** Initialisation
-  new, unsafeNew, replicate, replicateM, clone,
+  new, unsafeNew, replicate, replicateM, generate, generateM, clone,
 
   -- ** Growing
   grow, unsafeGrow,
@@ -619,6 +619,24 @@ replicate n x = stToPrim $ basicUnsafeReplicate (delay_inline max 0 n) x
 replicateM :: (PrimMonad m, MVector v a) => Int -> m a -> m (v (PrimState m) a)
 {-# INLINE replicateM #-}
 replicateM n m = munstream (MBundle.replicateM n m)
+
+-- | /O(n)/ Create a mutable vector of the given length (0 if the length is negative)
+-- and fill it with results of applying function to each index.
+generate :: (PrimMonad m, MVector v a) => Int -> (Int -> a) -> m (v (PrimState m) a)
+{-# INLINE generate #-}
+generate n f = stToPrim $ generateM n (return . f)
+
+-- | /O(n)/ Create a mutable vector of the given length (0 if the length is
+-- negative) and fill it with results of applying function to each
+-- index. Iteration is done string from 0 indexes.
+generateM :: (PrimMonad m, MVector v a) => Int -> (Int -> m a) -> m (v (PrimState m) a)
+{-# INLINE generateM #-}
+generateM n f = do
+  vec <- new n
+  let loop i | i >= n    = return vec
+             | otherwise = do unsafeWrite vec i =<< f i
+                              loop (i + 1)
+  loop 0
 
 -- | Create a copy of a mutable vector.
 clone :: (PrimMonad m, MVector v a) => v (PrimState m) a -> m (v (PrimState m) a)
--- a/Data/Vector/Generic/Mutable.hs
+++ b/Data/Vector/Generic/Mutable.hs
@@ -881,14 +881,14 @@ imapM_ :: (PrimMonad m, MVector v a) => (Int -> a -> m b) -> v (PrimState m) a -
 {-# INLINE imapM_ #-}
 imapM_ f v = forI_ v $ \i -> f i =<< unsafeRead v i
 
--- | /O(n)/ Apply  the monadic action to every element  of the vector,
--- discarding the results. It's same as @flip mapM_@.
+-- | /O(n)/ Apply the monadic action to every element of the vector,
+-- discarding the results. It's same as the @flip mapM_@.
 forM_ :: (PrimMonad m, MVector v a) => v (PrimState m) a -> (a -> m b) -> m ()
 {-# INLINE forM_ #-}
 forM_ = flip mapM_
 
 -- | /O(n)/ Apply the monadic action to every element of the vector
--- and its index, discarding the results. It's same as @flip imapM_@.
+-- and its index, discarding the results. It's same as the @flip imapM_@.
 iforM_ :: (PrimMonad m, MVector v a) => v (PrimState m) a -> (Int -> a -> m b) -> m ()
 {-# INLINE iforM_ #-}
 iforM_ = flip imapM_
--- a/Data/Vector/Generic/Mutable.hs
+++ b/Data/Vector/Generic/Mutable.hs
@@ -896,31 +896,31 @@ iforM_ = flip imapM_
 -- | /O(n)/ Pure left fold.
 foldl :: (PrimMonad m, MVector v a) => (b -> a -> b) -> b -> v (PrimState m) a -> m b
 {-# INLINE foldl #-}
-foldl f = ifoldl (const f)
+foldl f = ifoldl (\b _ -> f b)
 
 -- | /O(n)/ Pure left fold with strict accumulator.
 foldl' :: (PrimMonad m, MVector v a) => (b -> a -> b) -> b -> v (PrimState m) a -> m b
 {-# INLINE foldl' #-}
-foldl' f = ifoldl' (const f)
+foldl' f = ifoldl' (\b _ -> f b)
 
 -- | /O(n)/ Pure left fold (function applied to each element and its index).
-ifoldl :: (PrimMonad m, MVector v a) => (Int -> b -> a -> b) -> b -> v (PrimState m) a -> m b
+ifoldl :: (PrimMonad m, MVector v a) => (b -> Int -> a -> b) -> b -> v (PrimState m) a -> m b
 {-# INLINE ifoldl #-}
 ifoldl f b0 v = stToPrim $ loop 0 b0
   where
     loop i b | i >= n = return b
              | otherwise = do a <- unsafeRead v i
-                              loop (i + 1) $ f i b a
+                              loop (i + 1) $ f b i a
     n = length v
 
 -- | /O(n)/ Pure left fold with strict accumulator (function applied to each element and its index).
-ifoldl' :: (PrimMonad m, MVector v a) => (Int -> b -> a -> b) -> b -> v (PrimState m) a -> m b
+ifoldl' :: (PrimMonad m, MVector v a) => (b -> Int -> a -> b) -> b -> v (PrimState m) a -> m b
 {-# INLINE ifoldl' #-}
 ifoldl' f b0 v = stToPrim $ loop 0 b0
   where
     loop i !b | i >= n = return b
               | otherwise = do a <- unsafeRead v i
-                               loop (i + 1) $ f i b a
+                               loop (i + 1) $ f b i a
     n = length v
 
 -- | /O(n)/ Pure right fold.
--- a/Data/Vector/Generic/Mutable.hs
+++ b/Data/Vector/Generic/Mutable.hs
@@ -46,7 +46,9 @@ module Data.Vector.Generic.Mutable (
   -- * Folds
   mapM_, imapM_,
   foldl, foldl', foldM, foldM',
+  foldr, foldr', foldrM, foldrM',
   ifoldl, ifoldl', ifoldM, ifoldM',
+  ifoldr, ifoldr', ifoldrM, ifoldrM',
 
   -- * Modifying vectors
   nextPermutation,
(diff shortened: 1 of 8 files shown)


Bodigrim:
I have no issues with
```haskell
foldl :: (PrimMonad m, MVector v a) => (b -> a -> b) -> b -> v (PrimState m) a -> m b
```
But I'm unsure about monadic folds like 
```
foldM :: (PrimMonad m, MVector v a) => (b -> a -> m b) -> b -> v (PrimState m) a -> m b
```
Do we really need them? What useful monadic effect could happen in `b -> a -> m b`? It feels dangerous that this action can modify `v (PrimState m) a`, for example. I could probably contrive some examples, but not convinced that they embody enough justification to enshrine such idiom.

lehins:
`foldlM` is most useful of them all, because you can implement all other mutable iterators with it. Also note that it is not just `foldM`, because unlike lists, iterating monadically from the right is also useful and efficient. 

`ifoldlM` is essentially a for loop from imperative languages on steroids. It gives you index, value of the element, accumulator and ability to affect the world. It also gives you freedom to mutate the vector in-place as you go through it. I think it is the most underappreciated concept in Haskell because everyone tries really hard to live in the pure immutable world.

@Bodigrim You should not feel any danger about it, because there nothing unsafe about monadic folds.

On the other hand my opinion about `foldl` is quite the opposite, I think it sends the wrong message of false purity:
```
foldl :: (PrimMonad m, MVector v a) => (b -> a -> b) -> b -> v (PrimState m) a -> m b
foldl f acc vec = ..
```
It gives you this feeling that the function `f` is pure and thus the whole operation will have no effects, while in reality this vector can be mutated concurrently by another thread while this "foldl" is going through this mutable vector.

So I would recommend is adding all four:
```
foldlM :: (PrimMonad m, MVector v a) => (b -> a -> m b) -> b -> v (PrimState m) a -> m b
ifoldlM :: (PrimMonad m, MVector v a) => (b -> Int -> a -> m b) -> b -> v (PrimState m) a -> m b
foldrM :: (PrimMonad m, MVector v a) => (a -> b -> m b) -> b -> v (PrimState m) a -> m b
ifoldrM :: (PrimMonad m, MVector v a) => (Int -> a -> b -> m b) -> b -> v (PrimState m) a -> m b
```

Shimuuar:
I don't think possibility that callback function will modify vector is a problem. `mapM_` has same problem as any function taking monadic callback. Similarly imperative programmer can mutate vector in for loop. All these functions are basically for loop specialized in various ways


As for pure folds. I think about them as syntactic convenience. To avoid wrapping pure function into monads if it happens to be pure. Compare:

```haskell
foldl (+) 0 vec
foldM (\x y -> pure (x +y)) 0 vec
```
Also I don't see how they could be interpreted as pure. Result is wrapped in monad after all. And vectors are being modified concurrently all bets are off in any case.

P.S. Agree about right folds

lehins:
> I think about them as syntactic convenience

100% 

I am not against adding folds with pure functions. They definitely can't be interpreted as pure, so it is no surprise you don't see it ;) . All I am saying is that they can be _mistakenly_ interpreted as pure by a novice user, because by itself a `foldl` can't do any mutation. That doesn't make them less useful as syntactic convenience as you very well put it.

Data/Vector/Generic/Mutable.hs
```diff
@@ -774,6 +810,89 @@ unsafeExchange v i x = UNSAFE_CHECK(checkIndex) "unsafeExchange" i (length v)
                          unsafeWrite v i x
                          return y
 
+-- Folds
+-- -----
+
+forI_ :: (Monad m, MVector v a) => v (PrimState m) a -> (Int -> m b) -> m ()
```
konsumlamm, line 816: Shouldn't it be `PrimMonad m`?

Shimuuar: Not really. This function doesn't use anything from PrimMonad, Callback that passed as parameter usually does

konsumlamm: Isn't `PrimMonad m` required to use `PrimState m`? `PrimState` is defined inside the `PrimMonad` class, so I thought that constraint would be required to be able to use it.

Shimuuar: Sorry for late answer. Not quite. It's true that in order to define `PrimState` type family instance one have to define `PrimMonad` instance. But when it comes to use associated type work just like stand alone type families. Since we don't use any methods of `PrimMonad` we don't need to bring its dictionary here.

Shimuuar:
@lehins, @bodigrim any onjection about functions being added or about naming? If not I'll finish this PR. Added functions:

- generate/genrateM
- modifyM/unsafeModifyM
- mapM_, imapM_
- foldl, foldl', foldM, foldM'
- ifoldl, ifoldl', ifoldM, ifoldM'

lehins:
This sounds great. Will you add right folds in this PR too or postpone it to a later one?

Shimuuar:
I'll add them here. Thanks for reminding about them

Data/Vector/Generic/Mutable.hs
```diff
@@ -618,6 +623,24 @@ replicateM :: (PrimMonad m, MVector v a) => Int -> m a -> m (v (PrimState m) a)
 {-# INLINE replicateM #-}
 replicateM n m = munstream (MBundle.replicateM n m)
 
+-- | /O(n)/ Create a mutable vector of the given length (0 if the length is negative)
+-- and fill it with results of applying function to each index.
```
konsumlamm, line 627:
```suggestion
-- and fill it with the results of applying the function to each index.
```

konsumlamm:
If you're also adding right fold (I assume including monadic right folds), wouldn't it make more sense to use `foldlM` instead of `foldM`?

Data/Vector/Generic/Mutable.hs
```diff
@@ -618,6 +623,24 @@ replicateM :: (PrimMonad m, MVector v a) => Int -> m a -> m (v (PrimState m) a)
...
+generate :: (PrimMonad m, MVector v a) => Int -> (Int -> a) -> m (v (PrimState m) a)
+{-# INLINE generate #-}
+generate n f = stToPrim $ generateM n (return . f)
+
+-- | /O(n)/ Create a mutable vector of the given length (0 if the length is
+-- negative) and fill it with results of applying function to each
+-- index. Iteration is done string from 0 indexes.
```
konsumlamm, line 634:
```suggestion
-- negative) and fill it with the results of applying the monadic function to each
-- index. Iteration starts at index 0.
```

Data/Vector/Generic/Mutable.hs
```diff
@@ -718,6 +741,12 @@ modify :: (PrimMonad m, MVector v a) => v (PrimState m) a -> (a -> a) -> Int ->
 modify v f i = BOUNDS_CHECK(checkIndex) "modify" i (length v)
              $ unsafeModify v f i
 
+-- | Modify the element at the given position using a monadic function.
+modifyM :: (PrimMonad m, MVector v a) => v (PrimState m) a -> (a -> m a) -> Int -> m ()
+{-# INLINE modifyM #-}
+modifyM v f i = BOUNDS_CHECK(checkIndex) "modify" i (length v)
```
konsumlamm, line 747:
```suggestion
modifyM v f i = BOUNDS_CHECK(checkIndex) "modifyM" i (length v)
```

Data/Vector/Generic/Mutable.hs
```diff
@@ -774,6 +810,89 @@ unsafeExchange v i x = UNSAFE_CHECK(checkIndex) "unsafeExchange" i (length v)
...
+
+forI_ :: (Monad m, MVector v a) => v (PrimState m) a -> (Int -> m b) -> m ()
+{-# INLINE forI_ #-}
+forI_ v f = loop 0
+  where
+    loop i | i >= n    = return ()
+           | otherwise = f i >> loop (i + 1)
+    n = length v
+
+-- | /O(n)/ Apply monadic action to every element of vector discarding results.
```
konsumlamm, line 824:
```suggestion
-- | /O(n)/ Apply the monadic action to every element of the vector, discarding the results.
```

Data/Vector/Generic/Mutable.hs
```diff
@@ -774,6 +810,89 @@ unsafeExchange v i x = UNSAFE_CHECK(checkIndex) "unsafeExchange" i (length v)
...
+mapM_ :: (PrimMonad m, MVector v a) => (a -> m b) -> v (PrimState m) a -> m ()
+{-# INLINE mapM_ #-}
+mapM_ f v = forI_ v $ \i -> f =<< unsafeRead v i
+
+-- | /O(n)/ Apply monadic action to every element of vector and its index discarding results.
```
konsumlamm, line 829:
```suggestion
-- | /O(n)/ Apply the monadic action to every element of the vector and its index, discarding the results.
```

Data/Vector/Generic/Mutable.hs
```diff
@@ -774,6 +810,89 @@ unsafeExchange v i x = UNSAFE_CHECK(checkIndex) "unsafeExchange" i (length v)
...
+foldl :: (PrimMonad m, MVector v a) => (b -> a -> b) -> b -> v (PrimState m) a -> m b
+{-# INLINE foldl #-}
+foldl f = ifoldl (const f)
+
+-- | /O(n)/ Pure left fold with strict accumulator.
+foldl' :: (PrimMonad m, MVector v a) => (b -> a -> b) -> b -> v (PrimState m) a -> m b
+{-# INLINE foldl' #-}
+foldl' f = ifoldl' (const f)
+
+-- | /O(n)/ Pure left fold (function applied to each element and its index)
```
konsumlamm, line 844:
```suggestion
-- | /O(n)/ Pure left fold (function applied to each element and its index).
```

Data/Vector/Generic/Mutable.hs
```diff
@@ -774,6 +810,89 @@ unsafeExchange v i x = UNSAFE_CHECK(checkIndex) "unsafeExchange" i (length v)
...
+ifoldl :: (PrimMonad m, MVector v a) => (Int -> b -> a -> b) -> b -> v (PrimState m) a -> m b
+{-# INLINE ifoldl #-}
+ifoldl f b0 v = stToPrim $ loop 0 b0
+  where
+    loop i b | i >= n = return b
+             | otherwise = do a <- unsafeRead v i
+                              loop (i + 1) $ f i b a
+    n = length v
+
+-- | /O(n)/ Pure left fold with strict accumulator (function applied to each element and its index)
```
konsumlamm, line 854:
```suggestion
-- | /O(n)/ Pure left fold with strict accumulator (function applied to each element and its index).
```

Data/Vector/Generic/Mutable.hs
```diff
@@ -774,6 +810,89 @@ unsafeExchange v i x = UNSAFE_CHECK(checkIndex) "unsafeExchange" i (length v)
...
+
+-- | /O(n)/ Monadic fold.
+foldM :: (PrimMonad m, MVector v a) => (b -> a -> m b) -> b -> v (PrimState m) a -> m b
+{-# INLINE foldM #-}
+foldM f = ifoldM (const f)
+
+-- | /O(n)/ Monadic fold with strict accumulator.
+foldM' :: (PrimMonad m, MVector v a) => (b -> a -> m b) -> b -> v (PrimState m) a -> m b
+{-# INLINE foldM' #-}
+foldM' f = ifoldM (const f)
```
konsumlamm, line 872:
```suggestion
foldM' f = ifoldM' (const f)
```

Data/Vector/Generic/Mutable.hs
```diff
@@ -774,6 +810,89 @@ unsafeExchange v i x = UNSAFE_CHECK(checkIndex) "unsafeExchange" i (length v)
...
+
+-- | /O(n)/ Monadic fold (action applied to each element and its index)
```
konsumlamm, line 874:
```suggestion
-- | /O(n)/ Monadic fold (action applied to each element and its index).
```

Data/Vector/Generic/Mutable.hs
```diff
@@ -774,6 +810,89 @@ unsafeExchange v i x = UNSAFE_CHECK(checkIndex) "unsafeExchange" i (length v)
...
+ifoldM :: (PrimMonad m, MVector v a) => (Int -> b -> a -> m b) -> b -> v (PrimState m) a -> m b
+{-# INLINE ifoldM #-}
+ifoldM f b0 v = loop 0 b0
+  where
+    loop i b | i >= n = return b
+             | otherwise = do a <- unsafeRead v i
+                              loop (i + 1) =<< f i b a
+    n = length v
+
+-- | /O(n)/ Monadic fold with strict accumulator. (action applied to each element and its index)
```
konsumlamm, line 884:
```suggestion
-- | /O(n)/ Monadic fold with strict accumulator (action applied to each element and its index).
```

Data/Vector/Generic/Mutable.hs
```diff
@@ -751,6 +780,13 @@ unsafeModify v f i = UNSAFE_CHECK(checkIndex) "unsafeModify" i (length v)
                    $ basicUnsafeRead v i >>= \x ->
                      basicUnsafeWrite v i (f x)
 
+-- | Modify the element at the given position using a monadic
+-- function. No bounds checks are performed.
+unsafeModifyM :: (PrimMonad m, MVector v a) => v (PrimState m) a -> (a -> m a) -> Int -> m ()
+{-# INLINE unsafeModifyM #-}
+unsafeModifyM v f i = UNSAFE_CHECK(checkIndex) "unsafeModify" i (length v)
```
konsumlamm, line 787:
```suggestion
unsafeModifyM v f i = UNSAFE_CHECK(checkIndex) "unsafeModifyM" i (length v)
```

Shimuuar:
@konsumlamm Thanks!

Shimuuar:
Only unresolved point is naming: `foldM` vs `foldlM`. I think that we should go with `foldM` in order to keep naming within library consistent. Immutable vectors use `foldM` (and don't provide `foldrM`) so does `Control.Monad` as well.

konsumlamm:
> Only unresolved point is naming: `foldM` vs `foldlM`. I think that we should go with `foldM` in order to keep naming within library consistent. Immutable vectors use `foldM` (and don't provide `foldrM`) so does `Control.Monad` as well.

Perhaps we could do both and make `foldM` and alias for `foldlM` (that's what `Data.Vector.Fusion.Bundle.Monadic` and `Data.Vector.Fusion.Stream.Monadic` do fwiw)? One point in favor of using `foldlM` is that `Data.Foldable` provides `foldlM` and `foldrM` (and no `foldM`).

lehins:
I actually was gonna suggest the same thing of naming `foldlM` instead of `foldM` as well, but then double checked how those functions are named for immutable vectors and decided not to suggest this in the end. I believe consistency is way more important then personal preference.

I am also not in favor of adding `foldlM` as synonyms because there is too much overhead for very little benefit. That would mean adding `foldlM` and `ifoldlM` in 16 different modules (we'd have to add them for immutable for consistency too). This would mean extra 32 functions that have no other purpose but act as synonyms. I really don't think its worth it, so IMHO `foldM` and `ifoldM` will do just fine.

Shimuuar:
I've added reexports for (hopefully) everything. Also as it turns out we didn't have reexports for `exchange`

Data/Vector/Generic/Mutable.hs
```diff
@@ -822,6 +860,161 @@ unsafeExchange v i x = UNSAFE_CHECK(checkIndex) "unsafeExchange" i (length v)
...
+{-# INLINE mapM_ #-}
+mapM_ f v = forI_ v $ \i -> f =<< unsafeRead v i
+
+-- | /O(n)/ Apply the monadic action to every element of the vector and its index, discarding the results.
+imapM_ :: (PrimMonad m, MVector v a) => (Int -> a -> m b) -> v (PrimState m) a -> m ()
+{-# INLINE imapM_ #-}
+imapM_ f v = forI_ v $ \i -> f i =<< unsafeRead v i
+
+-- | /O(n)/ Apply  the monadic action to every element  of the vector,
+-- discarding the results. It's same as @flip mapM_@.
```
konsumlamm, line 885:
```suggestion
-- | /O(n)/ Apply the monadic action to every element of the vector,
-- discarding the results. It's the same as @flip mapM_@.
```
Ditto for the reexports.

Data/Vector/Generic/Mutable.hs
```diff
@@ -822,6 +860,161 @@ unsafeExchange v i x = UNSAFE_CHECK(checkIndex) "unsafeExchange" i (length v)
...
+forM_ :: (PrimMonad m, MVector v a) => v (PrimState m) a -> (a -> m b) -> m ()
+{-# INLINE forM_ #-}
+forM_ = flip mapM_
+
+-- | /O(n)/ Apply the monadic action to every element of the vector
+-- and its index, discarding the results. It's same as @flip imapM_@.
```
konsumlamm, line 891:
```suggestion
-- and its index, discarding the results. It's the same as @flip imapM_@.
```
Ditto for the reexports.

konsumlamm (changes requested):

Shimuuar:
Not ready for merge. I started writing tests and found out that I used wrong order of parameters for indexed folds. And now GHC produces tests suite that crashes. I'll start investigating this tomorrow

Shimuuar:
Turns out writing tests is pretty neat idea. They caught both discrepancies in API between pure and mutable versions and violations of memory safety. Now I think PR is ready for review.

tests/Utilities.hs
```diff
@@ -321,9 +330,15 @@ iscanr f z = scanr (uncurry f) z . zip [0..]
 ifoldr :: (Int -> a -> b -> b) -> b -> [a] -> b
 ifoldr f z = foldr (uncurry f) z . zip [0..]
 
-ifoldM :: Monad m => (a -> Int -> a -> m a) -> a -> [a] -> m a
+ifoldM :: Monad m => (b -> Int -> a -> m b) -> b -> [a] -> m b
 ifoldM = indexedLeftFold foldM
 
+foldrM :: Monad m => (a -> b -> m b) -> b -> [a] -> m b
+foldrM f z = foldr (\a mb -> f a =<< mb) (pure z)
```
lehins, line 337: How come we need to define it instead of using `Data.Foldable.foldrM` instead?

lehins (approved):
Just a few minor suggestions. Haven't found any correctness issues, so it looks great.

Data/Vector/Generic/Mutable.hs
```diff
@@ -620,6 +627,26 @@ replicateM :: (PrimMonad m, MVector v a) => Int -> m a -> m (v (PrimState m) a)
 {-# INLINE replicateM #-}
 replicateM n m = munstream (MBundle.replicateM n m)
 
+-- | /O(n)/ Create a mutable vector of the given length (0 if the length is negative)
+-- and fill it with the results of applying the function to each index.
+generate :: (PrimMonad m, MVector v a) => Int -> (Int -> a) -> m (v (PrimState m) a)
+{-# INLINE generate #-}
+generate n f = stToPrim $ generateM n (return . f)
```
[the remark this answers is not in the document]
Shimuuar: It allows GHC to work in the ST monad instead of whatever tower of monad transformers is thrown at it. GHC may or may not be able to optimize those binds. Hopefully this will reduce amount of work optimizer has to do and may even result in better code. Same for other function that take pure function as parameters

Data/Vector/Generic/Mutable.hs
```diff
@@ -822,6 +862,161 @@ unsafeExchange v i x = UNSAFE_CHECK(checkIndex) "unsafeExchange" i (length v)
...
+
+-- | /O(n)/ Pure left fold with strict accumulator (function applied to each element and its index).
+ifoldl' :: (PrimMonad m, MVector v a) => (b -> Int -> a -> b) -> b -> v (PrimState m) a -> m b
+{-# INLINE ifoldl' #-}
+ifoldl' f b0 v = stToPrim $ loop 0 b0
+  where
+    loop i !b | i >= n = return b
+              | otherwise = do a <- unsafeRead v i
+                               loop (i + 1) $ f b i a
+    n = length v
```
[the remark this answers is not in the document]
Shimuuar: Good point

tests/Utilities.hs
Shimuuar: I forgot about it :)

tests/Tests/Vector/Property.hs
```diff
@@ -470,6 +482,110 @@ testPolymorphicFunctions _ = $(testProperties [
...
+      = (\f z v -> Identity $ runST $ MV.foldM' (\b -> pure . runIdentity . f b) z =<< V.thaw v)
+      `eq` foldM
+    prop_mut_foldrM :: P ((a -> a -> Identity a) -> a -> v a -> Identity a)
+      = (\f z v -> Identity $ runST $ MV.foldrM (\a -> pure . runIdentity . f a) z =<< V.thaw v)
+      `eq`
+      foldrM
+    prop_mut_foldrM' :: P ((a -> a -> Identity a) -> a -> v a -> Identity a)
+      = (\f z v -> Identity $ runST $ MV.foldrM' (\a b -> pure $ runIdentity $ f a b) z =<< V.thaw v)
+      `eq`
+      foldrM
```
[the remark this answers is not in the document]
Shimuuar: We don't have lazy `foldrM` either. I'll add them in separate PR. This PR is large enough as it is

Shimuuar:
@lehins I finally addressed your suggestions. I also added changelog

lehins:
@Shimuuar This looks great. Changes are ready to be merged, but it would be nice to cleanup the commit history. If you aren't up to it I can just squash all of the changes into one and merge it.

Once it is merged I'll backport changes in this PR to vector-0.12.3.0 since we need to make another non-breaking release anyways that includes #372

Shimuuar:
Let just squash it. it not _that_ big and changes are sometimes interwoven and couldn't be cleaned up easily

Bodigrim (approved):

PR #339, opened 2020-11-07
Add since pragmas to every function added since last release

Shimuuar:
It's not clear what is right way to add since pragmas. Their values are decided
when release is made and not when function is added.

One possibility is to write `@since NEXT` and then sed it to correct value when doing release. One question is how to not forget it? @lehins @Bodigrim what is your opinion?

### DIFF
--- a/Data/Vector.hs
+++ b/Data/Vector.hs
@@ -724,7 +724,8 @@ unfoldrN :: Int -> (b -> Maybe (a, b)) -> b -> Vector a
 {-# INLINE unfoldrN #-}
 unfoldrN = G.unfoldrN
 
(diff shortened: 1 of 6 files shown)


Bodigrim:
Yeah, I do not know a good, low-maintenance solution for `@since` pragmas. That's why I usually do not bother with it. IMO it's better do not use `@since` at all than to risk applying it in a wrong, inconsistent or incomplete way.

Shimuuar:
On other hand these pragmas are great from user's PoV. On the third any attempt to add them manually require a lot of discipline from programmers and it's something programmers don't possess. On fourth they're fairly mechanical and in principle could be inserted by some tool. So question is why don't we have such tool? 

So yes this is poor approach. I'm closing this PR

PR #345, opened 2020-12-30
Support GHC 9.0

Bodigrim:
Tested with GHC 9.0 RC1. This is a bit speculative, because `doctest` and `gauge` do not build against GHC 9.0, but otherwise looks good. 

```
packages: vector.cabal
tests: true

allow-newer:
  *:base,
  *:Cabal,
  doctest:ghc

source-repository-package
  type: git
  location: https://github.com/dreixel/syb
```


Shimuuar:
@Bodigrim  please rebase on top of master so CI could pass and let just merge it.

PR #347, opened 2021-01-12
Use eager indexing for !?

treeowl:
* Make `(!?)` extract a value from the vector eagerly.
* Make the range check for `(!?)` use one comparison instead of two.

### DIFF
--- a/Data/Vector/Generic.hs
+++ b/Data/Vector/Generic.hs
@@ -241,8 +241,13 @@ infixl 9 !?
 -- | O(1) Safe indexing
 (!?) :: Vector v a => v a -> Int -> Maybe a
 {-# INLINE_FUSED (!?) #-}
-v !? i | i < 0 || i >= length v = Nothing
-       | otherwise              = Just $ unsafeIndex v i
+-- Lengths are never negative, so we can check  0 <= i < length v
+-- using one unsigned comparison.
+v !? i | (fromIntegral i :: Word) >= fromIntegral (length v)
+       = Nothing
+       | otherwise
+         -- Use basicUnsafeIndexM @Maybe to perform the indexing eagerly.
+       = basicUnsafeIndexM v i
 
 -- | /O(1)/ First element
 head :: Vector v a => v a -> a


Data/Vector/Generic.hs
```diff
@@ -241,8 +241,13 @@ infixl 9 !?
 -- | O(1) Safe indexing
 (!?) :: Vector v a => v a -> Int -> Maybe a
 {-# INLINE_FUSED (!?) #-}
-v !? i | i < 0 || i >= length v = Nothing
-       | otherwise              = Just $ unsafeIndex v i
+-- Lengths are never negative, so we can check  0 <= i < length v
+-- using one unsigned comparison.
```
phadej, line 245: I don't understand this comment and logic. Is the idea that 

```
fromIntegral (minBound :: Int) :: Word
9223372036854775808
```

is always bigger than any realistic length `Vector` could be?

I think that should explained more clearly, this is "too clever" otherwise.

treeowl: @phadej, right; a vector of length over `maxBound :: Int` will break assumptions all over the place. I can add a comment to that effect.

Bodigrim: Existing definition has a benefit of short-circuiting in compile time when `i` is a statically known negative number. But I do not think that this is an important scenario.

Shimuuar: It's probably worth adding note that `fromIntegral (i :: Int) :: Word > fromIntegral (maxBound :: Int) :: Word` for all negative Ints. Such optimization is not entirely obvious

Shimuuar:
`(!)` and `indexM` shoudl benefit from same optimization. Also are there any benchmarks that show benefit of this optimization?

treeowl:
> `(!)` and `indexM` shoudl benefit from same optimization. Also are there any benchmarks that show benefit of this optimization?

No idea. The benefits will be pretty sensitive to branch prediction load, I imagine.

Shimuuar:
Even more importantly is to make sure that this change doesn't make things worse because GHC does something stupid with casts. Even more interesting target for such optimization is `BOUNDS_CHECK` macro. It's used everywhere

Bodigrim:
To be honest, I do not quite imagine`(!?)` in a tight, performance-sensitive loop, which randomly asks for elements inside and outside of the range. It's interesting to apply this optimization to `BOUNDS_CHECK` and try to benchmark / look at Core / look at assembly.

As always, the smallest PRs catch the biggest scrutiny :) Could the range check optimization be moved to a separate PR?

PR #353, opened 2021-01-16
Disable doctests for ghc-8.10

lehins:


Shimuuar:
Just for sake of consistency. GHC < 8.6 is disabled in cabal file. We should either disable both in cabal file or both using CPP. I prefer cabal file out of acute dislike of CPP (cabal file is not much better though)

```
  -- GHC 8.10 fails to run doctests for some reason
  if impl(ghc >= 8.10) && impl(ghc < 8.11)
    buildable: False
```
That's how I did this in #335

lehins:
I do prefer to use CPP for solving this problem, despite that I share your dislike for CPP, only because `buildable: false` prevents targeting doctests directly:
```
$ stack test :vector-doctest --resolver nightly
Selected resolver: nightly-2021-01-15
The following components have 'buildable: False' set in the cabal configuration, and so cannot be targets:
    vector:test:vector-doctest
To resolve this, either provide flags such that these components are buildable, or only specify buildable targets.
$ echo $?
1
```

However since we don't target doctests directly in CI, the issue I described isn't relevant to us, so I did go ahead and cherry picked your commit

PR #354, opened 2021-01-16
Move all vector package related code into it's own directory. Fix #350

### LINKED ISSUE #350: Repository directory restructure
We need a top level package folder `vector`. This has been a common practice for years, so its time to get vector to catch up.

lehins:
@Shimuuar and @Bodigrim this is what I meant by restructure in #350 Reason why I'd really like this restructure is to identify which code belongs to which package. There are three separate projects here: vector, benchmarks and old-testsuite. And it is not immediately clear what `internal` and `include` folders are either and which project they belong to.

Also added `stack.yaml` file, to make it easier for contributors that use `stack`

Also removed benchmarks sources from `vector.cabal` file, they really don't need to be included into the package itself. If you think otherwise I can bring them back.

No worries about current PRs, git is quite smart about following renames like that and a simple rebase should do the right thing


Bodigrim:
How about nuking `old-testsuite`? Seems we do not run it at all, and I do not believe that it is still buildable after 11 years.

Bodigrim:
Dunno, such restructuring will make switching between commits extremely painful, especially if we intend to cherry-pick stuff back and forth. I do not feel that this pain is well justified. If we delete `old-testsuite`, we are left with only two packages, which are actually one package split for technical reasons. I'd rather keep it this way.

> Also removed benchmarks sources from vector.cabal file, they really don't need to be included into the package itself. 

Totally makes sense.

lehins:
> How about nuking old-testsuite?

I wouldn't rush with this, I just got old-testsuite/microsuite to work and it actually flagged some issues with fusion. Also I don't like dumping stuff, keeping it even for historical reasons is a good idea. I'll give it a shot and see if I can get the rest of the old suite to work and see what I can find. Even to check how good vector is with backwards compatibility is already a worthy experiment.

>  especially if we intend to cherry-pick stuff back and forth. I do not feel that this pain is well justified.

I don't understand what pain is this you are talking about? For example earlier I rebased this PR onto master that had quite a few changes without any problems or conflicts. I am telling you git is quite smart about moving files.

I believe into nicely organized repository, makes me wanna do more work on it. Just look how nice it is:

image

versus this!!

image

To me its totally worth it!

Bodigrim:
>  I just got old-testsuite/microsuite to work and it actually flagged some issues with fusion. 

Cool, even better. But in this case it can be made just a regular test suite, part of the `vector` package, right?

>  I am telling you git is quite smart about moving files.

`git` is, but what about `cabal` / `stack`? I believe they will trigger a full rebuild each time when you switch between commits.

My point is that semantically library, tests and benchmarks are the same package. Strategically I'd rather eliminate obstacles, which prevent them from being built as a single package (I have a patch for `gauge`, which allows this), than introduce explicit boundaries. 

To be honest, I do not see much difference between the screenshots. I like keeping `cabal` file and changelog at the top level, and I do not like to duplicate `README.md` both at top level and inside of `vector`. You also probably need to duplicate `LICENSE`, retaining a top-level copy, because it technically refers to all contents of the repo.

(On the other hand, this restructuring does not affect me personally in an adverse way)

Shimuuar:
Also should we split streams into separate package (#355) repository won't be single package anymore. In that context such change makes sense.

lehins:
>  But in this case it can be made just a regular test suite, part of the vector package, right?

`microsuite` is a bunch of files with `main` in them and each file is compiled by the test suite, so I don't know if it can be included into the vector's test suite, but maybe it could be. I don't know enough about it at the moment to tell for sure.

> I believe they will trigger a full rebuild each time when you switch between commits.

This will only be a temporary issue. Once we get vector-0.13 released with this restructure it will stop being a problem. Also I don't really see it a problem, more like a mere temporary inconvenience.

> My point is that semantically library, tests and benchmarks are the same package

I do agree with you, but with current tooling this is not always possible. While other people who work on those tools deal with obstacles, as a maintainer of a package not the build tools I prefer to work around such issues instead. For example I would be happy to have access to all the tests from vector package, this would allow me to test a custom instance of `MVector` and `Vector` classes I could have. Or provide Arbitrary instances for all user's of vector, would be awesome too. All of this, however, is a totally orthogonal issue to this restructure, so let's not worry about.

>  I do not like to duplicate README.md

Not a problem, I switched it to a symbolic link (I mistakenly create a hard link first) and did the same for LICENSE. Thi approach worked well for me in the past.

The biggest argument in favor of this restructure I have, as @Shimuuar well pointed out, switch of `Stream` into its own package will be another package in the repo, so top level cabal file won't as sensible any more. There is a reason why I brought up these two issues around the same time :wink:

Bodigrim:
> The biggest argument in favor of this restructure I have, as @Shimuuar well pointed out, switch of `Stream` into its own package will be another package in the repo, so top level cabal file won't as sensible any more. There is a reason why I brought up these two issues around the same time 😉

If you have mentioned this earlier, I could have saved my breath :) It totally makes sense if we separate streaming into its own package.

lehins:
@Bodigrim and @Shimuuar do you mind if I finally merge this? I wanna start on #355 and this one is a blocker.

Also [`vector-0.12.3.0`](https://hackage.haskell.org/package/vector-0.12.3.0) is out and I don't think we will be doing any more backports to 0.12.x Even if for some reason we'll have to, I'll get it done regardless of this structure divergence.

Bodigrim (approved):

Issue #331, opened 2020-09-21
minimumOn function


Boarders:
It would be nice to have a `minimumOn` function of the form:

```haskell
minimumOn :: (Ord b, Generic.Vector v a) => (a -> b) -> v a -> a 
```

A possible implementation could just adapt the existing one for `minimumBy`:
```haskell
minimumOn :: (Ord b, Vector v a) => (a -> b) -> v a -> a
{-# INLINE minimumOn #-}
minimumOn f = Bundle.foldl1' minBy . stream
  where
    {-# INLINE minBy #-}
    minBy x y = case compare (f x) (f y) of
                  GT -> y
                  _  -> x
```

lehins:
@Boarders I can see how such function can be useful (especially for `Storable`, `Unbox` and `Prim`, when `b` doesn't have an instance, while `a` does), but there is a problem with your suggested implementation. Namely function `f` will be evaluated `2*n - 1` times  and in case when `f` is really expensive this could be very noticeable.

A possibly better solution would be to first zip with that function over:

```haskell
minimumOn :: (Ord b, Vector v a) => (a -> b) -> v a -> a
{-# INLINE minimumOn #-}
minimumOn f = fst . Bundle.foldl1' minBy . Bundle.map (\a -> (a, f a)) . stream
  where
    {-# INLINE minBy #-}
    minBy x y = case compare (snd x) (snd y) of
                  GT -> y
                  _  -> x
```

If anyone is willing to implement this and `maximumOn` for all vector types with docs, tests and at least one benchmark, I would not be opposed to merging such addition.

Shimuuar:
Fixed by #356

PR #356, opened 2021-01-16
Add minimumOn and maximumOn functions (Issue #331)

claudi:
Implements the minimumOn and maximumOn function requested in the mentioned request, using the implementation described in the [reply](https://github.com/haskell/vector/issues/331#issuecomment-696248852). It also adds tests, documentation and benchmarks for the mentioned functions.

This is my first contribution to an open source project. Hopefully everything is alright.


Data/Vector.hs
```diff
@@ -1519,6 +1520,12 @@ maximumBy :: (a -> a -> Ordering) -> Vector a -> a
 {-# INLINE maximumBy #-}
 maximumBy = G.maximumBy
 
+-- | /O(n)/ Yield the maximum element of the vector according to the given
+-- comparison function. The vector may not be empty.
```
Bodigrim, line 1524: While I understand what you mean, it is misleading to call `a -> b` a "comparison function". It does not compare anything, right? I'd say "Yield an element, at which the given function takes the maximum values amongst all elements of the vector" or similar. 

Also it would be nice to mention how ties are resolved (e. g., arbitrary?).

benchmarks/Main.hs
```diff
@@ -61,4 +62,6 @@ main = do
     , bench "findIndexR" $ whnf findIndexR ((<indexFindThreshold), as)
     , bench "findIndexR_naïve" $ whnf findIndexR_naive ((<indexFindThreshold), as)
     , bench "findIndexR_manual" $ whnf findIndexR_manual ((<indexFindThreshold), as)
+    , bench "minimumOn"  $ whnf (minOn (\x -> x*x*x)) as
+    , bench "maximumOn"  $ whnf (maxOn (\x -> x*x*x)) as
```
Bodigrim, line 66: I'd benchmark `U.minimumBy` and `U.maximumBy` directly, without intermediaries in another module.

Bodigrim:
Thanks for tests and benchmarks :) You are very welcome to contribute.

Boarders:
Thanks for adding this @claudi! :)

Data/Vector.hs
claudi: > Also it would be nice to mention how ties are resolved (e. g., arbitrary? the left-most element? the right-most element?).

I forgot to add this in 536d57baa263c950d2e7854d40deb6ad1162d4a7, I will include it in a second.

lehins:
Way to go with a first contribution @claudi Keep up the good work ;)

Normally I'd also ask for an entry in changelog, but this time I'm gonna have to do it myself anyways, so no worries on it.

Shimuuar:
@lehins Too fast :) implementation doesn't match documentation

```
>>> maximumOn fst $ Data.Vector.fromList [(1,1), (1,2)]
(1,2)
>>> minimumOn fst $ Data.Vector.fromList [(1,1), (1,2)]
(1,1)
```

@claudi never mind about bug. I'm going to add doctest for that behavior. This is good case for tool for copying haddocks I made. (Hasn't published it yet)

claudi:
Should I fix the problem myself and submit another pull request?

Shimuuar:
@claudi please don't. Fix is trivial and I'm using this as test case for debugging tool for copying haddocks. 

Thanks for contribution!

lehins:
@Shimuuar Good catch.

PR #358, opened 2021-01-17
Add stack.yaml and get rid of compile time warnings

lehins:
Exactly what's in the title.


emilypi:
Builds fine on my side, no warnings!

PR #360, opened 2021-01-17
Improve haddock of `grow` functions and clarify their semantics. Fix #36

### LINKED ISSUE #36: Semantics of grow/unsafeGrow
I'd like to clarify the semantics of `grow` and `unsafeGrow`, and hopefully update the documentation appropriately. From looking at the code for boxed, unboxed, and storable vectors, it seems that in all cases, `grow`/`unsafeGrow` will create a new mutable vector and leave the original mutable vector unchanged. The upshot would be that it would be valid to modify the original vector without affecting the new vector, and vice-versa. Two questions:
1. Have I misread the code, and in fact that's _not_ the way the current three vector types work?
2. Do we want to codify this contract in the documentation as a promise for future behavior?

lehins:
Before anyone suggests that we might switch to resize primops in the future that are available from ghc-prim here is my answer:

* Those primops have exactly the same semantics: the allocate and copy when growing (shrinking is different however)
* If we ever get primops that can do inplace growing like `realloc` does, we better create new functions for such functionality. Switching semantics like that is too dangerous.

### DIFF
--- a/Data/Vector/Generic/Mutable/Base.hs
+++ b/Data/Vector/Generic/Mutable/Base.hs
@@ -87,8 +87,9 @@ class MVector v a where
                    -> v s a   -- ^ source
                    -> ST s ()
 
-  -- | Grow a vector by the given number of elements. This method should not be
-  -- called directly, use 'unsafeGrow' instead.
+  -- | Grow a vector by the given number of elements. Allocates a new vector and
+  -- copies all of the elements over starting at 0 index. This method should not
+  -- be called directly, use 'grow'\/'unsafeGrow' instead.
   basicUnsafeGrow  :: v s a -> Int -> ST s (v s a)
 
   {-# INLINE basicUnsafeReplicate #-}
--- a/Data/Vector/Mutable.hs
+++ b/Data/Vector/Mutable.hs
@@ -309,15 +309,49 @@ clone = G.clone
 -- Growing
 -- -------
 
--- | Grow a vector by the given number of elements. The number must be
--- positive.
+-- | Grow a boxed vector by the given number of elements. The number must be
+-- non-negative. Same semantics as in `G.grow` for generic vector. It differs
+-- from @grow@ functions for unpacked vectors, however, in that only pointers to
+-- values are copied over, therefore values themselves will be shared between
+-- two vectors. This is an important distinction to know about during memory
+-- usage analysis and in case when values themselves are of a mutable type, eg.
+-- `Data.IORef.IORef` or another mutable vector.
+--
+-- ====__Examples__
+--
+-- >>> import qualified Data.Vector as V
+-- >>> import qualified Data.Vector.Mutable as MV
+-- >>> mv <- V.thaw $ V.fromList ([10, 20, 30] :: [Integer])
+-- >>> mv' <- MV.grow mv 2
+--
+-- The two extra elements at the end of the newly allocated vector will be
+-- uninitialized and will result in an error if evaluated, so me must overwrite
+-- them with new values first:
+--
+-- >>> MV.write mv' 3 999
+-- >>> MV.write mv' 4 777
+-- >>> V.unsafeFreeze mv'
+-- [10,20,30,999,777]
+--
+-- It is important to note that the source mutable vector is not affected when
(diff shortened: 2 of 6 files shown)


Data/Vector/Generic/Mutable/Base.hs
```diff
@@ -87,8 +87,9 @@ class MVector v a where
                    -> v s a   -- ^ source
                    -> ST s ()
 
-  -- | Grow a vector by the given number of elements. This method should not be
-  -- called directly, use 'unsafeGrow' instead.
+  -- | Grow a vector by the given number of elements. Allocates a new vector and
+  -- copies all of the elements over starting at 0 index. This method should not
+  -- be called directly, use 'unsafeGrow' instead.
```
Shimuuar, line 92: I'd add reference to 'grow' as well.
```suggestion
  -- be called directly, use 'grow'\/'unsafeGrow' instead.
```

Shimuuar (commented):
> Before anyone suggests that we might switch to resize primops

I wanted to suggest to just describe existing semantics. people depends on it whether it's official or not. 

Other than minor nitpick I think PR's good.

Data/Vector/Mutable.hs
```diff
@@ -309,15 +309,49 @@ clone = G.clone
 -- Growing
 -- -------
 
--- | Grow a vector by the given number of elements. The number must be
--- positive.
+-- | Grow a boxed vector by the given number of elements. The number must be
+-- non-negative. Same semantics as in `G.grow` for generic vector. It differes
```
Bodigrim, line 313:
```suggestion
-- non-negative. Same semantics as in `G.grow` for generic vector. It differs
```

Data/Vector/Storable/Mutable.hs
```diff
@@ -371,15 +371,49 @@ clone = G.clone
...
+--
+-- ====__Examples__
+--
+-- >>> import qualified Data.Vector.Storable as VS
+-- >>> import qualified Data.Vector.Storable.Mutable as MVS
+-- >>> mv <- VS.thaw $ VS.fromList ([10, 20, 30] :: [Int])
+-- >>> mv' <- MVS.grow mv 2
+--
+-- Extra elements at the end of the newly allocated vector are initialized to
+-- default value, which for storable vector will always be 0. However, if
```
Bodigrim, line 385: Not all `Storable` types are numerics (`Bool`, `Char`, `()`, etc.). Same for `Primitive` above.

Data/Vector/Unboxed/Mutable.hs
```diff
@@ -187,15 +187,49 @@ clone = G.clone
...
+--
+-- ====__Examples__
+--
+-- >>> import qualified Data.Vector.Unboxed as VU
+-- >>> import qualified Data.Vector.Unboxed.Mutable as MVU
+-- >>> mv <- VU.thaw $ VU.fromList ([('a', 10), ('b', 20), ('c', 30)] :: [(Char, Int)])
+-- >>> mv' <- MVU.grow mv 2
+--
+-- Extra elements at the end of the newly allocated vector are initialized to default
+-- value, which for primitive vector will always be some form of 0. However, if
```
Bodigrim, line 201:
```suggestion
-- value, which for unboxed vector will always be some form of 0. However, if
```

Data/Vector/Storable/Mutable.hs
lehins: Sure, but it doesn't change the fact that memory will contain zeros.

lehins: If you have better wording for this, I am all ears.

Bodigrim: How about "Extra memory is initialized with zero bytes" or similar, without refering to elements? There are quite non-trivial instances of `Storable` in the wild, for which zeros does not mean anything.

lehins: @Bodigrim I did rephrase it a bit, hopefully it makes things more clear. This sort of explanation ideally would be explained in haddock for `new`, but one thing at a time.

Bodigrim (approved):

Shimuuar (approved):

Issue #362, opened 2021-01-18
Semantics for maximumOn & friends


Shimuuar:
This is followup to #356. I stumbled on it while writing doctests. in #180 maximumBy definition was changed. 

#### Data.List & vector0.13
- `minimumBy` returns first tie
- `maximumBy` returns last tie
#### vector-0.12
- `minimumBy` returns first tie
- `maximumBy` returns first tie

maximumOn should I think match behavior of  maximumBy. Question is whether we should match functions for lists which aren't very consistent or should revert to 0.12

P.S. Data.Foldable.maximumBy function definition doesn't make any promises on ties

Bodigrim:
I think we should clearly document existing behaviour of all four functions, but do not modify `maximumOn` to follow deprecated behaviour of `maximumBy`. Doing otherwise will just increase an amount of potential breakage in 0.13.

Shimuuar:
I don't understand what do you mean by this. `maximumOn` & maximumBy` should behave identically otherwise we'll have two function that do same thing but slightly differently.

Real question is whether we should keep change to `maximumBy` from return first to return last, or should we revert to definition from 0.12 (return first).

Bodigrim:
> Real question is whether we should keep change to maximumBy from return first to return last, or should we revert to definition from 0.12 (return first).

I'm in favor of reverting #180. 

`base` does not specify tie behaviour of `minimumBy` / `maximumBy`, and theoretically can change it in any major release (which happens pretty often). We should not be chase compatibility with it entirely at our own expense. #180 introduced a dangerous silent breaking change, and I do not think that the motivation was strong enough.

Shimuuar:
Looks like consensus is to revert and document behavior. I'll update #364 accordingly

PR #363, opened 2021-01-24
Add some doctests

Shimuuar:
They're mostly copied from #101 with exception of maximum/minimum related ones (see #362, they're to be fixed separately). Also examples from accum & accumulate are converted to doctests

### DIFF
--- a/Data/Vector.hs
+++ b/Data/Vector.hs
@@ -956,7 +956,11 @@ unsafeUpdate_ = G.unsafeUpdate_
 -- | /O(m+n)/ For each pair @(i,b)@ from the list, replace the vector element
 -- @a@ at position @i@ by @f a b@.
 --
--- > accum (+) <5,9,2> [(2,4),(1,6),(0,3),(1,7)] = <5+3, 9+6+7, 2+4>
+-- ==== __Examples__
+--
+-- >>> import qualified Data.Vector as V
+-- >>> V.accum (+) (V.fromList [1000.0,2000.0,3000.0]) [(2,4),(1,6),(0,3),(1,10)]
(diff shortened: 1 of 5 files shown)


lehins:
Taking a quick peek at it looks great. One comment I have right now is that it is important to stay consistent with respect to doctests because users will mimic them in their code. In this case in particular I am talking about qualified imports. I've personally almost always used these:

```haskell
import qualified Data.Vector.Mutable as MV
import qualified Data.Vector.Generic as VG
import qualified Data.Vector.Generic.Mutable as MVG
import qualified Data.Vector.Storable as VS
import qualified Data.Vector.Storable.Mutable as MVS
import qualified Data.Vector.Primitive as VP
import qualified Data.Vector.Primitive.Mutable as MVP
import qualified Data.Vector.Unboxed as VU
import qualified Data.Vector.Unboxed.Mutable as MVU
```

Those were also the names I've used in recently added PR #360 

I am not terribly set on those names, but keeping them consistent I think is very important. However I really don't like  the  `as S`  for storable vector, because often `import qualified Foreign.Storable as S` also shows up in scope,

Shimuuar:
I always used G/V/P/U for qualified imports. But VG/V/VP/VU is just as good to me. I'll update doctests

lehins:
@Shimuuar If you are busy I can update those qualified imports. I'd, like to merge this PR, just so I can included these doctests in the `vector-0.12.2.0` release I'd like to cut today

Shimuuar:
@lehins Done.

lehins:
@Shimuuar Nice, thank you. I'll  merge it when CI completes

Bodigrim (approved):

Issue #214, opened 2018-05-16
Document that uniq keeps the last duplicate


rightfold:
The documentation of _uniq_ does not specify exactly which elements are kept.

```haskell
-- assume a /= b, b == c
Vector.uniq [a, b, c] = [a, c]
```

This is important when _Eq_ instances do not consider all parts of the data structure.

Do you think this should be documented (and tested), or should this be left unstable?

andrewthad:
If it is unstable, this should at least be documented.

glguy:
I'd prefer things like this be documented to be unstable so that people aren't expected to memorize which behavior functions like these have when they're reading code using them. Instead of supporting the use of a surprise `Eq` instance via functions like `uniq` functions that were explicit about their behavior and were parameterized over a relation that didn't have the implicit baggage of `Eq` would be much more appropriate.

PR #364, opened 2021-01-26
Add doctests for minmum/maximum related functions

### LINKED ISSUE #362: Semantics for maximumOn & friends

Shimuuar:
Doctests assume current semantics as it's implemented in code. For minimum first tie
    wins, for maximum last one,
    
    I think most convincing reason to keep this semantics is that maximum/minimum
    pair works in exactly same way when equal values could be distinguished:
    
      >>> data F = F Int Char deriving Show
      >>> instance Eq F  where F i _ == F j _ = i == j
      >>> instance Ord F where compare = comparing (\(F i _) -> i)
      >>> V.maximum [F 1 'a', F 1 'b']
      F 1 'b'
      >>> V.minimum [F 1 'a', F 1 'b']
      F 1 'a'
    
    As a bonus it matches behavior of list functions

@lehins @Bodigrim what's your opinion on matter? 

Fixes #362

P.S. @lehins you may want to pick second commit for 0.12.2.

### DIFF
--- a/Data/Vector.hs
+++ b/Data/Vector.hs
@@ -1638,15 +1638,33 @@ maximum :: Ord a => Vector a -> a
 {-# INLINE maximum #-}
 maximum = G.maximum
 
--- | /O(n)/ Yield the maximum element of the vector according to the given
--- comparison function. The vector may not be empty.
+-- | /O(n)/ Yield the maximum element of the vector according to the
+-- given comparison function. The vector may not be empty. In case of
+-- a tie the last occurrence wins.
+--
+-- ==== __Examples__
+--
+-- >>> import Data.Ord
+-- >>> import qualified Data.Vector as V
+-- >>> V.maximumBy (comparing fst) $ V.fromList [(2.0,'a'), (1.0,'b')]
+-- (2.0,'a')
+-- >>> V.maximumBy (comparing fst) $ V.fromList [(1.0,'a'), (1.0,'b')]
+-- (1.0,'b')
 maximumBy :: (a -> a -> Ordering) -> Vector a -> a
 {-# INLINE maximumBy #-}
 maximumBy = G.maximumBy
 
 -- | /O(n)/ Yield the maximum element of the vector by comparing the results
--- of a key function on each element. In case of a tie, the first occurrence
+-- of a key function on each element. In case of a tie, the last occurrence
 -- wins. The vector may not be empty.
+--
+-- ==== __Examples__
+--
+-- >>> import qualified Data.Vector as V
+-- >>> V.maximumOn fst $ V.fromList [(2.0,'a'), (1.0,'b')]
+-- (2.0,'a')
+-- >>> V.maximumOn fst $ V.fromList [(1.0,'a'), (1.0,'b')]
+-- (1.0,'b')
 maximumOn :: Ord b => (a -> b) -> Vector a -> a
 {-# INLINE maximumOn #-}
 maximumOn = G.maximumOn
@@ -1657,8 +1675,18 @@ minimum :: Ord a => Vector a -> a
 {-# INLINE minimum #-}
 minimum = G.minimum
 
--- | /O(n)/ Yield the minimum element of the vector according to the given
--- comparison function. The vector may not be empty.
+-- | /O(n)/ Yield the minimum element of the vector according to the
+-- given comparison function. The vector may not be empty. In case of
+-- a tie, the first occurrence wins. The vector may not be empty.
+--
+-- ==== __Examples__
+--
+-- >>> import Data.Ord
+-- >>> import qualified Data.Vector as V
+-- >>> V.minimumBy (comparing fst) $ V.fromList [(2.0,'a'), (1.0,'b')]
+-- (1.0,'b')
+-- >>> V.minimumBy (comparing fst) $ V.fromList [(1.0,'a'), (1.0,'b')]
+-- (1.0,'a')
 minimumBy :: (a -> a -> Ordering) -> Vector a -> a
 {-# INLINE minimumBy #-}
 minimumBy = G.minimumBy
@@ -1666,6 +1694,14 @@ minimumBy = G.minimumBy
 -- | /O(n)/ Yield the minimum element of the vector by comparing the results
 -- of a key function on each element. In case of a tie, the first occurrence
 -- wins. The vector may not be empty.
+--
+-- ==== __Examples__
+--
+-- >>> import qualified Data.Vector as V
+-- >>> V.minimumOn fst $ V.fromList [(2.0,'a'), (1.0,'b')]
+-- (1.0,'b')
+-- >>> V.minimumOn fst $ V.fromList [(1.0,'a'), (1.0,'b')]
+-- (1.0,'a')
 minimumOn :: Ord b => (a -> b) -> Vector a -> a
 {-# INLINE minimumOn #-}
 minimumOn = G.minimumOn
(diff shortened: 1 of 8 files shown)


lehins:
I don't know about this. Semigroup instance for `Max` and `Min` have different behavior:

```haskell
λ> import Data.Semigroup
λ> import qualified Data.List.NonEmpty as NE
λ> getMax $ sconcat $ fmap Max $ (Arg 1 'a') NE.:| [Arg 1 'b' :: Arg Int Char]
Arg 1 'a'
λ> getMin $ sconcat $ fmap Min $ (Arg 1 'a') NE.:| [Arg 1 'b' :: Arg Int Char]
Arg 1 'a'
```

Data/Vector.hs
```diff
@@ -1530,38 +1530,86 @@ product = G.product
 
 -- | /O(n)/ Yield the maximum element of the vector. The vector may not be
 -- empty.
+--
+-- ==== __Examples__
+--
+-- >>> import qualified Data.Vector as V
+-- >>> V.maximum $ V.fromList [2.0, 1.0]
```
lehins, line 1537: Minor improvement to show that `maximum` for generic vectors works with other type of vectors:
```suggestion
-- >>> import qualified Data.Vector.Generic as VG
-- >>> import qualified Data.Vector as V
-- >>> VG.maximum $ V.fromList ([2.0, 1.0] :: [Integer])
-- >>> import qualified Data.Vector as VS
-- >>> VG.maximum $ VS.fromList ([1.0, 2.0, 3.0] :: [Int])
```

Shimuuar:
You're right. List functions behave in weird way:

```haskell
>>> minimum   [Arg 1 'a', Arg 1 'b']
Arg 1 'a'
>>> minimumBy compare   [Arg 1 'a', Arg 1 'b']
Arg 1 'a'

>>> maximum   [Arg 1 'a', Arg 1 'b']
Arg 1 'a'
>>> maximumBy  compare   [Arg 1 'a', Arg 1 'b']
Arg 1 'b'
```

master behaves in same way, and 0.12 always returns first element. Maybe reverting changes to maximum in name of reducing breakage is better... But I don't have strong opinion here.

Shimuuar:
I've added few more one which uses doctests: D.V.Generic.maximum, examples for maximum/minimum which involve `Arg` and specified `uniq` behavior (#214)

Wrt maximum/minimum we have following choice:

#### Don't change code in master. 
1. Same behavior as list analogs
2. Breaking change for maximumBy/maxIndexBy
3. It's possible to get either return first or last behavior for both maximum/minimum using Down. But that's a bit of hack

#### Revert #180
1. No breaking changes
2. All functions work consistently

I'm slightly lean towards keeping things as they're in master

Data/Vector.hs
Shimuuar: I think it would be better to explain this in top-level documentation of modules. Otherwise some example will show this difference and some won't.

Shimuuar:
I've reverted #180 as was discussed in #362 and added notice about difference with `Data.List.maximumBy`. I think PR is finally ready to merge

Data/Vector/Generic.hs
```diff
@@ -1778,42 +1787,96 @@ product :: (Vector v a, Num a) => v a -> a
...
+-- Arg 1.0 'a'
 minimum :: (Vector v a, Ord a) => v a -> a
 {-# INLINE minimum #-}
 minimum = Bundle.foldl1' min . stream
 
--- | /O(n)/ Yield the minimum element of the vector according to the given
--- comparison function. The vector may not be empty.
+-- | /O(n)/ Yield the minimum element of the vector according to the
+-- given comparison function. The vector may not be empty. In case of
+-- a tie, the first occurrence wins. The vector may not be empty.
```
Bodigrim, line 1870: This says "The vector may not be empty" twice. 
On a separate note, I suggest a stronger wording "The vector must be non-empty".

Data/Vector/Primitive.hs
```diff
@@ -1354,13 +1360,24 @@ product :: (Prim a, Num a) => Vector a -> a
...
 maximum :: (Prim a, Ord a) => Vector a -> a
 {-# INLINE maximum #-}
 maximum = G.maximum
 
--- | /O(n)/ Yield the maximum element of the vector according to the given
--- comparison function. The vector may not be empty.
+-- | /O(n)/ Yield the maximum element of the vector according to the
+-- given comparison function. The vector may not be empty. In case of
+-- a tie the first occurrence wins. This behavior is different from
+-- 'Data.List.maximumBy' which returns last tie.
```
Bodigrim, line 1380:
```suggestion
-- 'Data.List.maximumBy' which returns the last tie.
```

tests/Tests/Vector/Property.hs
```diff
@@ -541,8 +541,8 @@ testOrdFunctions _ = $(testProperties
 listMaxIndexFMW :: Ord a => [a] -> Int
 listMaxIndexFMW  = ( fst  . extractFMW .  sconcat . DLE.fromList . fmap FMW . zip [0 :: Int ..])
 
-listMaxIndexLMW :: Ord a => [a] -> Int
-listMaxIndexLMW = ( fst  . extractLMW .  sconcat . DLE.fromList . fmap LMW . zip [0 :: Int ..])
+-- listMaxIndexLMW :: Ord a => [a] -> Int
+-- listMaxIndexLMW = ( fst  . extractLMW .  sconcat . DLE.fromList . fmap LMW . zip [0 :: Int ..])
```
Bodigrim, line 545: Why is it left commented out?

Data/Vector/Primitive.hs
Shimuuar: Fixed thanks

Data/Vector/Generic.hs
Shimuuar: "may not be empty" is used elsewhere in documentation. Do you think we should change them too?

tests/Tests/Vector/Property.hs
Shimuuar: Not used any more since we don't have function that picks last tie. I commented in out to keep it around. Just in case

Data/Vector/Generic.hs
Bodigrim: I'm in favor to change it everything, but not necessarily in this PR.

Data/Vector/Unboxed.hs
```diff
@@ -1392,29 +1401,85 @@ product :: (Unbox a, Num a) => Vector a -> a
 product = G.product
 
 -- | /O(n)/ Yield the maximum element of the vector. The vector may not be
--- empty.
+-- empty. In a case of a tie the first occurrence wins.
+--
+-- ==== __Examples__
+--
+-- >>> import qualified Data.Vector.Unboxed as VU
+-- >>> import qualified Data.Vector.Generic as VG
```
lehins, line 1409:
```suggestion
```

Data/Vector/Unboxed.hs
```diff
@@ -1392,29 +1401,85 @@ product :: (Unbox a, Num a) => Vector a -> a
...
+-- >>> VU.maximum $ VU.fromList [2.0, 1.0]
+-- 2.0
+-- >>> VG.maximum $ VU.fromList [2.0, 1.0]
+-- 2.0
```
lehins, line 1413:
```suggestion
```
Seems like a copy paste error, right?

tests/Tests/Vector/Property.hs
lehins: I think its best to remove dead code like this, it can be later recovered easily from `listMaxIndexFMW` if necessary.

lehins (approved):
Looks great!

tests/Tests/Vector/Property.hs
Shimuuar: Removed

Data/Vector/Generic.hs
Shimuuar: Lets do it separately. I fixed double empty

Data/Vector/Unboxed.hs
Shimuuar: Possibly. I don't want to dig through commit to find out, Fixed

Shimuuar:
I've addressed everything I thing and squashed few commits together

Bodigrim (approved):
Nice!

PR #368, opened 2021-02-05
Switch benchmarks from gauge to tasty-bench

Bodigrim:
This is a continuation of #325. 

The main motivation is to make benchmarks an inherent part of `vector` package, simplifying maintenance. No third party packages are rebuilt for benchmarks. We already use `tasty` for the test suite, so the only additional package is `tasty-bench` itself. 

Another motivation is that I'm frustrated with `gauge` still throwing `index out of bounds` two years after this was fixed in `statistics` (see https://github.com/vincenthz/hs-gauge/issues/88). It seems that in the long run inlining whole packages is not very prudent.

I restored command-options to specify vector size and random seed, because now they come free of charge: we reuse `tasty` internal mechanisms. CC @AndreasPK


.github/workflows/ci.yml
```diff
@@ -83,3 +83,7 @@ jobs:
     - name: Test
       run: |
         cabal test all
+    # ----------------
+    - name: Bench
+      run: |
+        cabal bench all --benchmark-options='--list-tests'
```
Shimuuar, line 89: Running benchmarks on the cloud I think is pure waste of CPU time. Too many jobs running on same hardware, too many layers of virtualization. I think it's enough to check that benchmarks build successfully. `Build` step already do that

Shimuuar (commented):
Being able to run benchmarks without rebuilding world is great!

.github/workflows/ci.yml
Bodigrim: I think it is useful at least to list benchmarks. 
Updated with tweaked `cabal configure` options, to avoid redundant reconfiguration.

Bodigrim:
@Shimuuar @lehins this is ready for review.

lehins:
@Bodigrim Give me a couple of days, I am really busy right now. I wanna try this PR first before merging it.

Shimuuar:
LGTM

lehins:
Looks great. 
@Bodigrim, awesome job on tasty-bench. Happy to see a lightweight alternative to criterion and gauge.

PR #369, opened 2021-03-05
Add GHC 9.0 to CI

Shimuuar:
And that's it

### DIFF
--- a/vector.cabal
+++ b/vector.cabal
@@ -41,7 +41,7 @@ Tested-With:
   GHC == 8.4.4,
   GHC == 8.6.5,
   GHC == 8.8.4,
-  GHC == 8.10.3
+  GHC == 8.10.4
   GHC == 9.0.1
 
 Cabal-Version:  >=1.10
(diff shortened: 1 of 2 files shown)


Shimuuar:
Not as smooth as I hoped:

    Turns out it requires cabal 3.4 to build and it's only available on ubuntu ATM
    so no 9.0 on macOS and windows. Hopefully this will be fixed at some point.
    
    Doctests don't work on GHC 9.0 as well as on 8.10. We need to fix that
    
    Also bump dependency on doctests. Things seem to work well.

vector.cabal
```diff
@@ -40,9 +40,9 @@ Tested-With:
   GHC == 8.2.2,
   GHC == 8.4.4,
   GHC == 8.6.5,
-  GHC == 8.8.1,
-  GHC == 8.10.1
-
+  GHC == 8.8.4,
+  GHC == 8.10.3
```
Bodigrim, line 44: Why not 8.10.4?

Shimuuar: Because CI matrix uses 8.10.3. But good point I should update this

Bodigrim (approved):

PR #378, opened 2021-04-10
Add iso-deriving for Unboxed instances

Shimuuar:
Another take on use of DerivingVia for defining Unbox instances. This should largely subsume TH-based deriving from th-vector-unbox. Verbosity is about same and there's no TH which is frequently breaks with each GHC release

P.S. It's still too verbose. I think use of data family was a mistake. They couldn't be derived and have to be defined for each type. On top of it we only have handful of possible representations. Type family + newtype wrapper for injectivity would've worked out much better.

### DIFF
--- a/vector/src/Data/Vector/Unboxed/Base.hs
+++ b/vector/src/Data/Vector/Unboxed/Base.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE DefaultSignatures #-}
+{-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE BangPatterns, CPP, MultiParamTypeClasses, TypeFamilies, FlexibleContexts #-}
 {-# LANGUAGE DeriveDataTypeable, StandaloneDeriving #-}
 {-# LANGUAGE PolyKinds #-}
@@ -53,6 +55,8 @@ import Data.Semigroup (Min(..),Max(..),First(..),Last(..),WrappedMonoid(..),Arg(
 import Data.Typeable ( Typeable )
 import Data.Data     ( Data(..) )
 import GHC.Exts      ( Down(..) )
+import GHC.Generics
+import Data.Coerce
 
 -- Data.Vector.Internal.Check is unused
 #define NOT_VECTOR_MODULE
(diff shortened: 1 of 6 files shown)


vector/src/Data/Vector/Unboxed/Base.hs
```diff
@@ -253,6 +254,86 @@ instance P.Prim a => G.Vector Vector (UnboxViaPrim a) where
   basicUnsafeCopy (MV_UnboxViaPrim mv) (V_UnboxViaPrim v) = G.basicUnsafeCopy mv v
   elemseq _ = seq
 
+-- | Isomorphism between type @a@ and its representation in unboxed
+-- vector @b@.
+class Isomorphic a b where
```
lehins, line 259: Considering that this is designed specifically for `Unbox` class, I can suggest two things:

* Add `Unbox` as a constraint for `b`: 
```haskell
class Unbox b => Isomorphic a b where
```
* Change name to something like `UnboxIso` or `IsoUnbox`

lehins (approved):
I love it!

It is similar to what I was suggesting here: https://github.com/haskell/vector/issues/315#issuecomment-643148132 but better

vector/src/Data/Vector/Unboxed/Base.hs
```diff
@@ -253,6 +254,86 @@ instance P.Prim a => G.Vector Vector (UnboxViaPrim a) where
...
+  -- | Convert value into it representation in unboxed vector.
+  toURepr   :: a -> b
+  -- | Convert value representation in unboxed vector back to value.
+  fromURepr :: b -> a
```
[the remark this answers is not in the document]
Shimuuar: I wonder are there cases when one could coerce but not GND? I need to think about this. Obvious downside it prevents other possible implementations

Also most important use case should be converting between product types and tuples. I'll try to cook up something using generics to see whether its possible

vector/src/Data/Vector/Unboxed/Base.hs
klapaucius: Is it right to make this class only for unboxed vectors though? What if someone wants partially unbox ```data Foo = Foo Int Int Double Text``` as some sort of a hybrid vector of three unboxed vectors and one boxed?

Shimuuar: As long as it's used with `As` newtype `Unbox` constraint on `b` is necessary: 
```haskell
instance (Isomorphic a b, Unbox b) => M.MVector MVector (As a b) where
instance (Isomorphic a b, Unbox b) => G.Vector Vector (As a b) where
```
It's of course possible to add more newtypes. For example one for sneaking boxed representations:

```haskell
import qualified Data.Vector.Mutable as MV

newtype Boxed a = Boxed a
newtype instance Vector (Boxed a) = V_Box (V.Vector a)
newtype instance MVector s (Boxed a) = MV_Box (MV.Vector s a)
{- necessary instances -}

data Foo = Foo Int Text

instance VU.Isomorphic Foo (Int,Boxed Text) where
   toURepr (Foo i s) = (i,Boxed a)
   fromURepr (i,Boxed a) = Foo i (Boxed a)
{- rest of boilerplate -}
```

Shimuuar: I tried adding superclass constraint `class Unbox b => IsoUnbox a b where` but decided against it. It requires adding `Unbox` constraint when defining `IsoUnbox` instances for polymorphic types:

```haskell
instance Unbox a => VU.IsoUnbox (Foo a) (Int,a) where
-- vs
instance VU.IsoUnbox (Foo a) (Int,a) where
```

It only creates busywork for users and `Vector`/`MVector` instances need Unbox constraint either way

vector/src/Data/Vector/Unboxed/Base.hs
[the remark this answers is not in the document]
Shimuuar: I started writing pessimistic reply about how we can't have good default instance... And then come up with interesting idea. We should coerce between generic representations and not between values themselves. As and illustration:
```
λ> data Foo = Foo Int Char deriving (Show,Generic)
λ> to . coerce . from $ Foo 1 'c' :: (Int,Char)
(1,'c')
λ> to . coerce . from $ Foo 1 'c' :: (Sum Int,Char)
(Sum {getSum = 1},'c')
λ> to . coerce . from $ (Sum (1::Int), 'c') :: Foo
Foo 1 'c'
```
It's even able to coerce between fields! This is definition I propose (sans INLINE):
```haskell
class IsoUnbox a b where
  -- | Convert value into it representation in unboxed vector.
  toURepr   :: a -> b
  default toURepr :: (Generic a, Generic b, Coercible (Rep a ()) (Rep b ())) => a -> b
  toURepr = to . idU . coerce . idU . from
  -- | Convert value representation in unboxed vector back to value.
  fromURepr :: b -> a
  default fromURepr :: (Generic a, Generic b, Coercible (Rep b ()) (Rep a ())) => b -> a
  fromURepr = to . idU . coerce . idU . from

idU :: f () -> f ()
idU = id
```

Only question is whether GHC guarantees that data types with same shape will have same shape. Think `(a :*: b) :*: c` vs `a :*: (b :*: c)`. I think they do but documentation is silent about this.

Shimuuar:
I added generic-based instance for `IsoUnbox`. It looks very nice. Only question is performance. I think it should be fine since we don't do anything except calling `to` and `from`

Issue #139, opened 2016-10-03
Docs for `unsafeThaw` should warn about the same things that `unsafePerformIO`


bitonic:
Everything that the docs for `unsafePerformIO` say regarding `NOINLINE`, `-fno-cse`, and `-fno-full-laziness` applies to `unsafeThaw`. In fact, I think it's even easier to fall into weird behavior with `unsafeThaw` -- for example this program is unsafe:

```
main :: IO ()
main = foo >> foo

foo :: IO ()
foo = do
  indexVector <- V.unsafeThaw $ V.generate n id
  x <- VM.read indexVector 5
  VM.write indexVector 5 (x * x)
  print x

{-
Will probably print
5
5
5
25
-}
```

(See also https://ghc.haskell.org/trac/ghc/ticket/12656#comment:9 for related woes)

bitonic:
Proposed new haddocks for `unsafeThaw`:

```
-- | /O(1)/ Unsafely convert an immutable vector to a mutable one without
-- copying. The immutable vector may not be used after this operation.
--
-- Moreover, @NOINLINE@, @-fno-cse@ and @-fno-full-laziness@ should be used to
-- ensure safety. See the documentation for 'unsafePerformIO' for more
-- information on this subject.
```

I'm currently unclear on whether `unsafeFreeze` needs the same warnings.

cartazio:
Hrmm. This is a good start!
I'm not sure if you need all three at the same time.
Or at least there's some style tricks to make it better.
Do we want to have the full exposition somewhere ?

What are the analogous docs / exposition for unsafe perform IO

aren't the safety requirements more like unsafe dupable perform IO for
unsafe thaw and unsafe freeze? Either way it gives you the same value wrt
array read and writes .  Just maybe a different heap allocated wrapper
constructor.

mgsloan:
> I'm currently unclear on whether `unsafeFreeze` needs the same warnings.

I don't think it does.  If you are sure that you do no further mutation of the mutable buffer, it doesn't matter if freezing it gets inlined.

dolio:
[...]

PR #381, opened 2021-04-19
docs: Explain that `unsafeThaw` allows reuse of argument on read-only ops

nh2:
This makes clear that usages such as

    unsafeCopy mutableTarget =<< unsafeThaw immutableSource

are safe.

Justification thanks to @lehins:

    It is safe to use it like that. unsafeThaw/unsafeFreeze only do
    a type cast for ByteArray/MutableByteArray.
    For boxed arrays it treats it a bit differently because GC needs
    to traverse mutable arrays, but that does not make this use case
    you mentioned unsafe in any way.

### DIFF
--- a/vector/src/Data/Vector.hs
+++ b/vector/src/Data/Vector.hs
@@ -2075,7 +2075,7 @@ unsafeFreeze :: PrimMonad m => MVector (PrimState m) a -> m (Vector a)
 unsafeFreeze = G.unsafeFreeze
 
 -- | /O(1)/ Unsafely convert an immutable vector to a mutable one without
--- copying. The immutable vector may not be used after this operation.
+-- copying. The immutable vector may not be used after this operation if the thawed vector is mutated.
 unsafeThaw :: PrimMonad m => Vector a -> m (MVector (PrimState m) a)
 {-# INLINE unsafeThaw #-}
 unsafeThaw = G.unsafeThaw
--- a/vector/src/Data/Vector/Generic.hs
+++ b/vector/src/Data/Vector/Generic.hs
@@ -2246,7 +2246,7 @@ freeze :: (PrimMonad m, Vector v a) => Mutable v (PrimState m) a -> m (v a)
 freeze mv = unsafeFreeze =<< M.clone mv
 
 -- | /O(1)/ Unsafely convert an immutable vector to a mutable one without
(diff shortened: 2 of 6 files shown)


lehins (approved):
:+1: Thanks fort he PR. I'll merge it tomorrow, just wanna give other guys a chance to comment.

vector/src/Data/Vector.hs
```diff
@@ -2075,7 +2075,7 @@ unsafeFreeze :: PrimMonad m => MVector (PrimState m) a -> m (Vector a)
 unsafeFreeze = G.unsafeFreeze
 
 -- | /O(1)/ Unsafely convert an immutable vector to a mutable one without
--- copying. The immutable vector may not be used after this operation.
+-- copying. The immutable vector may not be used after this operation if the thawed vector is mutated.
```
Bodigrim, line 2078: This double-conditioned sentence is quite entangled to parse. But besides stylistic nitpicking, I think it's valuable to explain "why?", what is the causal relationship here? I suggest something along the following lines:

```
Any mutation of the result (e. g., 'write') corrupts the input vector, breaking the promise of referential transparency. Thus the input must not be accessed after 'unsafeThaw', unless the output is used solely for reading (e. g., by 'copy').
```

nh2: I don't follow where the double-condition is in my sentence (there's only one `if` in it), but I do like your additional explanation.

Shimuuar:
Thanks for PR! Looks good. I propose to add following to `unsafeThaw` in order to fix #139

> Note that this is very dangerous function since GHC could easily introduce unwanted sharing by let-floating pure vector to the top level or destroy it by inlining. See documentation of 'unsafePerformIO' for discussion of this problem

lehins:
> I propose to add following to unsafeThaw in order to fix

Let's add this in a separate PR because right now it is not clear to me what recommendation to put in documentation for unsafeThaw with regards to its unsafety. Whatever is in unsafePerformIO documentation does not seem to be applicable exactly, see my example in ghc tracker: https://gitlab.haskell.org/ghc/ghc/-/issues/12656#note_347996

Shimuuar:
> Let's add this in a separate PR

Fair enough

Shimuuar:
Superseded by #386

PR #382, opened 2021-04-24
Legalize scanl1/scanr1 over an empty vector

gksato:
Now `scanl1`, `scanl1'`, `scanr1`, `scanr1'` gives an empty vector for an
empty input.
The definitions of these functions working on vectors,
bundles and streams, the documentations attached to these
definitions, and the related tests are correspondingly modified.

In the former definition, the functions `scanl1` and `scanr1` refused to
process empty vector and simply errorred.
However these functions could, and now they do, simply return an empty
vector in case of the absence of elements, as the corresponding
functions in Data.List in base do.
With this modification,
`postscanl' (<>) mempty`
can be rewritten to more generally applicable
`scanl1' (<>)`.

### DIFF
--- a/vector/changelog.md
+++ b/vector/changelog.md
@@ -35,6 +35,14 @@
      `Data.Vector.Storable{.Mutable}` to allow this (the onus is on the user
      to ensure that no `Storable` invariants are broken when using these
      functions).
+ * The functions `scanl1`, `scanl1'`, `scanr1`, and `scanr1'` for immutable
+   vectors are now defined when given empty vectors as arguments,
+   in which case they return empty vectors. This new behavior is consistent
+   with the one of the corresponding functions in `Data.List`.
+   Prior to this change, applying an empty vector to any of those functions
+   resulted in an error. This change was introduced in:
+   [#382](https://github.com/haskell/vector/pull/382)
+
 
 # Changes in version 0.12.3.0
 
--- a/vector/src/Data/Vector.hs
+++ b/vector/src/Data/Vector.hs
@@ -1929,17 +1929,37 @@ iscanl' :: (Int -> a -> b -> a) -> a -> Vector b -> Vector a
 {-# INLINE iscanl' #-}
 iscanl' = G.iscanl'
 
--- | /O(n)/ Scan over a non-empty vector
+-- | /O(n)/ Initial-value free scan over a vector
 --
 -- > scanl f <x1,...,xn> = <y1,...,yn>
 -- >   where y1 = x1
 -- >         yi = f y(i-1) xi
 --
+-- Note: Since 0.13, application of this to an empty vector no longer
+-- results in an error; instead produces an empty vector.
+--
+-- ==== __Examples__
+-- >>> import qualified Data.Vector as V
+-- >>> V.scanl1 min $ V.fromListN 5 [4,2,4,1,3 :: Int]
+-- [4,2,2,1,1]
+-- >>> V.scanl1 max $ V.fromListN 5 [1,3,2,5,4 :: Int]
+-- [1,3,3,5,5]
+-- >>> V.scanl1 min (V.empty :: V.Vector Int)
+-- []
 scanl1 :: (a -> a -> a) -> Vector a -> Vector a
 {-# INLINE scanl1 #-}
 scanl1 = G.scanl1
 
--- | /O(n)/ Scan over a non-empty vector with a strict accumulator
+-- | /O(n)/ Initial-value free scan over a vector with a strict accumulator
+--
+-- ==== __Examples__
+-- >>> import qualified Data.Vector as V
+-- >>> V.scanl1' min $ V.fromListN 5 [4,2,4,1,3 :: Int]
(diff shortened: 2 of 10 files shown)


gksato:
I'm looking through the failed checks, and the result looks so cryptic to me. macos-latest/8.6.5 tests fails in its doctest here, as does in my MacBook w/ ghc 8.6.5.
However the same test fails for current `master`(6b024a043bf7bb11fed2d2ddb39cd6a87183674c) on my macBook with the same error message, but it seems like it doesn't in here...

lehins:
macOS+ghc-8.6 error:

> vector-doctest: user error (Language.Haskell.GhciWrapper.close: Interpreter exited with an error (ExitFailure (-10)))

I've seen this error many times in my other project and a few times in vector now. It looks like a bug either in doctests or in ghc and I have a gut feeling it has something to do with mutable boxed vectors used in doctests. In other words we need to investigate this issue, but it is unlikely that this PR is at fault here.

ghc-9.0+Windows error:

> cabal.exe: Failed to build test:vector-tests-O0 from vector-0.13.0.1. The
failure occurred during the configure step.

The other problem with ghc-9 is a build error that comes from cabal and I am not sure what's at fault here

Shimuuar:
Yes defining `scanl1 f [] = []` makes sense. 

As for doctests. They are very flaky. I'll try to figure out why they're failing

gksato:
Thank you for your comments. I'm skimming through the logs of the checks, which all failed, from the second run triggered by my mistaken closure and reopen of this PR. I'm not at all familiar with GitHub CI and I've never read a single line of its document, but it looks as if `make sdist`s had failed because the `*.cabal` files, or maybe whole package, had been absent due to the lack of the run of `actions/checkout@v2`.  Maybe the following line is at blame: https://github.com/haskell/vector/blob/6b024a043bf7bb11fed2d2ddb39cd6a87183674c/.github/workflows/ci.yml#L48

Does it need to be fixed or not? Do we need a separate issue? I mean, re-testing is just a waste of computing resource if no new commit is added, but it is confusing the check mark is always replaced with cross mark when the author of the PR accidentally closes and reopens it.

lehins:
@Bodigrim Do you know why this line ^ is there? It doesn't make sense to me to not checkout code on CI

Shimuuar:
I think it was copy-pasted from somewhere without understanding what does it do

Shimuuar:
@gksato could you please rebase PR on top of current master and force push. That should trigger CI rebuild which will hopefully work

gksato:
@Shimuuar Thank you for your work. I think I've done it. Let's see what happens...

gksato:
Hooray, it worked!

Shimuuar:
Could you add changelog entry? And note that prior to 0.13 calling them on empty vector resulted in error.

P.S. I wrote tool for copying documentation from Generic module: https://github.com/Shimuuar/vector-doctweak It doesn't always work but saves a lot of work

P.P.S. If you add doctest examples it would be absolutely great

gksato:
Alrighty, I'll get it done today

gksato:
*Yuck*. I'm seeing some CI failure again. I've done it, anyway.

Shimuuar:
Sorry for the delay. I've been terribly busy as of late. 

For the record CI failures are:

#### ubuntu GHC7.10

Weird linking errors:

> Linking /home/runner/work/vector/vector/dist-newstyle/build/x86_64-linux/ghc-7.10.3/vector-0.13.0.1/t/vector-tests-O0/build/vector-tests-O0/vector-tests-O0 ...
> /usr/bin/ld: /home/runner/work/vector/vector/dist-newstyle/build/x86_64-linux/ghc-7.10.3/vector-0.13.0.1/t/vector-tests-O0/build/vector-tests-O0/vector-tests-O0-tmp/Main.o: relocation R_X86_64_32S against symbol `stg_upd_frame_info' can not be used when making a PIE object; recompile with -fPIE

And it's over 9 thousand lines about relocation R_X86_64_32S 

#### Ubuntu GHC 8.6.5

Doctests fail with:

> vector-doctest: /opt/ghc/8.6.5/bin/ghc-8.6.5: getPermissions:getFileStatus: does not exist (No such file or directory)

#### Ubuntu GHC 8.6.5

Same here

> vector-doctest: /opt/ghc/8.8.4/bin/ghc-8.8.4: getPermissions:getFileStatus: does not exist (No such file or directory)

Shimuuar:
I've restarted CI build and now everything works except for GHC7.10 which fails with linker error. I'm not sure what causes this problem and what to do with it

Other that that PR is good to go.

gksato:
Thanks for inspecting. It looks we have the same error in the current HEAD. The library itself builds, but the linker fails to link the test executables. How bothering.

Shimuuar:
CI also stands for constant irritation. I think I've fixed it. Could you please rebase on top of master again. Hopefully it will work this time

gksato:
Thanks for your work. It passed.

Shimuuar:
Finally! Thank you for your work and patience

Issue #383, opened 2021-04-26
Haddock incompatibility with older GHC


ag-eitilt:
I like to locally test my packages against every version the Hackage matrix would be testing, just to have the most accurate data before actually releasing, and several times have run into issues with `vector` by the time I get back to GHC-8.4.4 -- namely, the Haddock fails to build.  Since that's not marked in the `vector.cabal` (and since `cabal.project` files don't support `if()`), I need to manually add a `--constraint=vector<0.12.2` every time I try to test an older GHC, which I never remember to do until after flailing around for a bit.  The matrix (and presumably most users) don't have this issue because everything works fine if the Haddock is never built, but I have my global Cabal config set to build the docs for *everything*.

The issue seems to be with the `{-# UNPACK #-}` pragmas on `Data.Vector.Mutable.MVector`, and their interaction with the postfix Haddock; there might be other similar issues, but that's the first error the build hits.  It's too late at night for me to pick it apart at the moment, but `vector-0.12.2.0` added (admittedly needed) explanation of each argument to the constructor in a way which the older Haddock doesn't seem to be able to parse.  I'll look into this a bit further soon, but I just wanted to put this out there before I lost track of it again, and in case anyone else knows the fix off the top of their head.

ag-eitilt:
This might be the same issue as haskell/haddock#836, in which case it seems like there might not be much of a fix beyond hiding the docs behind an `#if MIN_VERSION_base(4,12,0)` and providing an `#else` without the additional Haddock.  `CPP` is already enabled for that module, so at least it won't be adding system complexity as well as the redundant code.

lehins:
Question is: do we really need to support old buggy versions of `haddock`? What does it buy us?

I think it is possible to use newer haddock version with older ghc, so if someone really needs to build documentation with ghc-8.4 and older they should be able to figure it out. Or maybe am I missing something?

Bodigrim:
@ag-eitilt you can put a constraint in a global cabal config. Or skip running haddock for old GHCs. Hackage Matrix does not run it, so why would you?

I don’t think this qualifies as an issue on vector side in 2021.

lehins:
I am 100% in agreement with @Bodigrim 
Closing as won't fix.

ag-eitilt:
[...]

PR #384, opened 2021-04-28
Implement a workaround for failing older version of haddock

lehins:
This is a fix for #383 that doesn't rely on CPP and a fix for #385


Shimuuar (approved):
I like that workaround!

PR #386, opened 2021-05-01
Expand documentation for unsafeThaw

### LINKED ISSUE #139: Docs for `unsafeThaw` should warn about the same things that `unsafePerformIO`

Shimuuar:
It now explains dangers of unsafeThaw. Fixes #139, supersedes #381, since it explains that read only use is safe

### DIFF
--- a/vector/src/Data/Vector.hs
+++ b/vector/src/Data/Vector.hs
@@ -2074,8 +2074,32 @@ unsafeFreeze :: PrimMonad m => MVector (PrimState m) a -> m (Vector a)
 {-# INLINE unsafeFreeze #-}
 unsafeFreeze = G.unsafeFreeze
 
--- | /O(1)/ Unsafely convert an immutable vector to a mutable one without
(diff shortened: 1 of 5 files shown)


Shimuuar:
No one complained about documentation PR for 3 weeks. So merging

PR #387, opened 2021-05-01
Change allocation strategy for unfoldrN & add warning for fromListN

### LINKED ISSUE #301: unfoldrN, unfoldrNM and fromListN are dangerous
Size argument supplied to all three functions `unfoldrN`, `unfoldrNM` and `fromListN` serves as  hint for upper bound for the length of a vector. The problem is that regardless of how big a vector would have been, the supplied upper bound memory is allocated, even if it too big, which can result in an application being killed with an asynchronous exception `HeapOverflow`

#### `fromListN`.hs:

```haskell
import qualified Data.Vector.Primitive as V

main = do
  print $ V.fromListN (maxBound `div` 8) xs
```

```shell
$ ghc fromListN.hs && ./fromListN
[1 of 1] Compiling Main             ( fromListN.hs, fromListN.o )
fromListN: Out of memory
```

#### `unfoldrN`.hs

```haskell
import qualified Data.Vector.Primitive as V

main = print (V.unfoldrN (maxBound `div` 8) (const Nothing) () :: V.Vector Int)
```

```shell
$ ghc unfoldrN.hs && ./unfoldrN
[1 of 1] Compiling Main             ( unfoldrN.hs, unfoldrN.o )
Linking unfoldrN ...
unfoldrN: Out of memory
```

#### `unfoldrNM`.hs

```haskell
import Control.Exception
import qualified Data.Vector.Primitive as V

main = do
  eRes <- try (V.unfoldrNM (maxBound `div` 8) (const $ pure Nothing) () :: IO (V.Vector Int))
  print (eRes :: Either SomeException (V.Vector Int))
```

```shell
$ stack exec -- ghc unfoldrNM.hs -O2 && ./unfoldrNM
[1 of 1] Compiling Main             ( unfoldrNM.hs, unfoldrNM.o )
Linking unfoldrNM ...
Left heap overflow
````

### PR DESCRIPTION
**Pull request by Shimuuar**
This PR changes allocation strategy for `unfoldrN{,M}` to `Unknown` and adds warning about possible DoS for fromListN. Here is reasoning from #301

> I think that changing type hint from Max to Unknown for unfoldrN/unfoldrNM is right thing to do. In addition to possible heap overflow it's poor strategy in cases when upper limit is used as some sort of safeguard and vector is usually much smaller. It just wastes memory.
>
> On other hand such change for fromListN makes no sense. It's mostly an optimization for cases when vector size is known in advance. It allows avoid reallocation of buffers when growing vector. Yes. it allows to induce heap overflow if size if hands of attacker. Without such preallocation we could just drop fromListN or implement it as fromList . take n. I think it's better to leave function as it is and just update documentation explaining possible dangers

@lehins, what's you opinion? 

Fixes #301

### DIFF
--- a/vector/src/Data/Vector/Generic.hs
+++ b/vector/src/Data/Vector/Generic.hs
@@ -2297,7 +2297,11 @@ fromList :: Vector v a => [a] -> v a
 {-# INLINE fromList #-}
 fromList = unstream . Bundle.fromList
 
--- | /O(n)/ Convert the first @n@ elements of a list to a vector.
+-- | /O(n)/ Convert the first @n@ elements of a list to a vector. It's
+-- expected that supplied list will be exactly @n@ elements long. As
+-- optimization this function allocates buffer for @n@ elements and
+-- could be used to DoS by exhausting memory if attacker controls that
+-- parameter.
 --
 -- @
 -- fromListN n xs = 'fromList' ('take' n xs)
--- a/vector/src/Data/Vector/Generic.hs
+++ b/vector/src/Data/Vector/Generic.hs
@@ -2298,10 +2298,10 @@ fromList :: Vector v a => [a] -> v a
 fromList = unstream . Bundle.fromList
 
 -- | /O(n)/ Convert the first @n@ elements of a list to a vector. It's
--- expected that supplied list will be exactly @n@ elements long. As
--- optimization this function allocates buffer for @n@ elements and
--- could be used to DoS by exhausting memory if attacker controls that
--- parameter.
+-- expected that the supplied list will be exactly @n@ elements long. As
+-- an optimization, this function allocates a buffer for @n@ elements, which
+-- could be used for DoS-attacks by exhausting the memory if an attacker controls
+-- that parameter.
 --
 -- @
 -- fromListN n xs = 'fromList' ('take' n xs)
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by Shimuuar** [comment]
After [discussion](https://github.com/haskell/vector/issues/301#issuecomment-845587128)  in #301 I think that problem is deeper. Why `Max` and `Exact` hints are treated alike? (#388)

**Comment by lehins** [comment]
I like this change.

@Shimuuar, do you mind if we include it as-is into 0.13 and get the deeper problem with bounds tackled in the future major release? if you can rebase this on master we can merge it in. If you are busy I can do it myself next week, so let me know if you oppose changes in this PR for any reason.

**Comment by Shimuuar** [comment]
Not at all. I'll rebase this PR today

File: vector/src/Data/Vector/Generic.hs [hunk 1]
```diff
@@ -2205,7 +2205,11 @@ fromList :: Vector v a => [a] -> v a
 {-# INLINE fromList #-}
 fromList = unstream . Bundle.fromList
 
--- | /O(n)/ Convert the first @n@ elements of a list to a vector
+-- | /O(n)/ Convert the first @n@ elements of a list to a vector. It's
+-- expected that supplied list will be exactly @n@ elements long. As
+-- optimization this function allocates buffer for @n@ elements and
+-- could be used to DoS by exhausting memory if attacker controls that
+-- parameter.
```
konsumlamm, line 2212:
```suggestion
-- | /O(n)/ Convert the first @n@ elements of a list to a vector. It's
-- expected that the supplied list will be exactly @n@ elements long. As
-- an optimization, this function allocates a buffer for @n@ elements, which
-- could be used for DoS-attacks by exhausting the memory if an attacker controls
-- that parameter.
```
(same for the other comments)

PR #389, opened 2021-05-23
Add inspection tests to vector

Shimuuar:
At this point its mostly skeleton with single test. At the moment I'm a bit at loss what to add. Maybe copy tests from text?

Tests are placed into separate test suite because common test suite is built with  both `-O0` and `-O2` and former will break fusion.

### DIFF
--- a/vector/tests-inspect/Inspect.hs
+++ b/vector/tests-inspect/Inspect.hs
@@ -1,0 +1,11 @@
+-- |
+{-# OPTIONS_GHC -fplugin=Test.Tasty.Inspection.Plugin #-}
+module Inspect where
+
+import Test.Tasty
+import qualified Data.Vector as V
+
+tests :: TestTree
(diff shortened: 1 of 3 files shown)


Bodigrim (approved):

Bodigrim:
Fusion tests in `text` were generated using my fork of `inspection-testing`: https://github.com/Bodigrim/inspection-testing/tree/6b99b613b30fc2216f391db08b857a54352e4f48/examples/text-api 
It should be possible to adapt the generator to `vector`.

Inspection testing relies on GHC plugins, and requires GHC 8.0.

Shimuuar:
@Bodigrim thanks for the input

Shimuuar:
I want to have working inspection tests for iso-deriving PR (#378) so I'm merging this one. Proper tests should be left for another PR

PR #399, opened 2021-08-09
Clarify kinds of As arguments explicitly

Bodigrim:
Otherwise GHC HEAD cannot figure out that `b` is restricted to be of kind `Type` and complains 
```
src/Data/Vector/Unboxed/Base.hs:345:61: error:
    • Expected a type, but ‘b’ has kind ‘k’
      ‘k’ is a rigid type variable bound by
        the data constructor ‘MV_UnboxAs’
        at src/Data/Vector/Unboxed/Base.hs:345:1-62
    • In the second argument of ‘MVector’, namely ‘b’
      In the type ‘(MVector s b)’
      In the definition of data constructor ‘MV_UnboxAs’
    |
345 | newtype instance MVector s (As a b) = MV_UnboxAs (MVector s b)
    |
```

This is because 
> Kind inference for data/newtype instance declarations is slightly more restrictive than before. See the user manual Kind inference for data/newtype instance declarations. This is a breaking change, albeit a fairly obscure one that corrects a specification bug.
> http://downloads.haskell.org/ghc/9.2.1-rc1/docs/html/users_guide/9.2.1-notes.html#language

### DIFF
--- a/vector/src/Data/Vector/Generic/Base.hs
+++ b/vector/src/Data/Vector/Generic/Base.hs
@@ -27,10 +27,11 @@ import qualified Data.Vector.Generic.Mutable.Base as M
 import           Data.Vector.Fusion.Util (Box(..), liftBox)
 
 import Control.Monad.ST
+import Data.Kind (Type)
 
 -- | @Mutable v s a@ is the mutable version of the immutable vector type @v a@ with
 -- the state token @s@. It is injective on GHC 8 and newer.
-type family Mutable (v :: * -> *) = (mv :: * -> * -> *) | mv -> v
+type family Mutable (v :: Type -> Type) = (mv :: Type -> Type -> Type) | mv -> v
 
 -- | Class of immutable vectors. Every immutable vector is associated with its
 -- mutable version through the 'Mutable' type family. Methods of this class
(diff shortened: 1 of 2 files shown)


Shimuuar (approved):
LGTM. Maybe it's time to start adding 9.2 to CI?

P.S. CPP is ugly

Bodigrim:
We can add `'9.2.0.20210422'` to a build matrix, but I'd rather wait for an official RC. 

With regards to CPP, my vote is to drop 7.10. FWIW neither `bytestring` nor `text` support it any longer.

lehins:
I am in favor of dropping 7.10. And yeah, let's wait for RC for 9.2, there is no need to rush on it.

Shimuuar:
+1 on dropping 7.10

lehins:
@Bodigrim Do you wanna fix those stars as well in this PR? GHC 9 yields a warning about those

```
vector/src/Data/Vector/Generic/Base.hs:type family Mutable (v :: * -> *) = (mv :: * -> * -> *) | mv -> v
vector/src/Data/Vector/Generic/Mutable.hs:type family Immutable (v :: * -> * -> *) :: * -> *
```

Bodigrim:
@lehins done.

lehins (approved):
Nice!

PR #401, opened 2021-08-10
Add `unsafeCast` to `Primitive` vectors

lehins:
While reviewing #398 I noticed that there is no `unsafeCast` for primitive vector.
 
We can't provide a cast that changes primitive vectors with elements of different sizes in the way that `Data.Vector.Storable.unsafeCast` does it. For example `Vector Word8` -> `Vector Word16` cast cannot be implemented with current `primitive`. However, this doesn't mean a cast like `Vector Word8` -> Word Int8` or a `Vector Float` -> `Vector Word32` cannot be useful for some reason.

### DIFF
--- a/vector/changelog.md
+++ b/vector/changelog.md
@@ -43,6 +43,7 @@
    resulted in an error. This change was introduced in:
    [#382](https://github.com/haskell/vector/pull/382)
 * Remove redundant `Storable` constraints on to/from `ForeignPtr` conversions
+* Add `unsafeCast` to `Primitive` vectors
 
 # Changes in version 0.12.3.0
 
--- a/vector/src/Data/Vector/Primitive.hs
+++ b/vector/src/Data/Vector/Primitive.hs
@@ -145,7 +145,7 @@ module Data.Vector.Primitive (
   toList, fromList, fromListN,
 
   -- ** Other vector types
-  G.convert,
+  G.convert, unsafeCast,
   unsafeCoerceVector,
 
   -- ** Mutable vectors
@@ -157,6 +157,7 @@ module Data.Vector.Primitive (
 
 import qualified Data.Vector.Generic           as G
 import           Data.Vector.Primitive.Mutable ( MVector(..) )
+import           Data.Vector.Internal.Check
 import qualified Data.Vector.Fusion.Bundle as Bundle
 import           Data.Primitive.ByteArray
 import           Data.Primitive ( Prim, sizeOf )
@@ -203,7 +204,7 @@ type role Vector nominal
 -- This is marginally safer than 'unsafeCast', since this function imposes an
 -- extra 'Coercible' constraint. This function is still not safe, however,
 -- since it cannot guarantee that the two types have memory-compatible
--- 'Storable' instances.
+-- 'Prim' instances.
 --
 -- Note that this function is unsafe. The @Coercible@ constraint guarantees
 -- that the element types are representationally equal. It however cannot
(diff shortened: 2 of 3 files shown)


Bodigrim (approved):

Shimuuar:
Or unsafeCast over newtypes for that matter

qrpnxz:
> We can't provide a cast that changes primitive vectors with elements of different sizes in the way that Data.Vector.Storable.unsafeCast does it.

Why not?

qrpnxz:
How unfortunate. Thanks for explaining.

qrpnxz:
A pinned `Primitive` variant could do it by implementing `*byteOff :: MutableByteArray# -> Int# -> State# s -> _` helper functions with `byteArrayContents#`, `plusAddr#`, and `*OffAddr#`:

```hs
```

qrpnxz:
A pinned `Primitive` variant could do it by implementing `*byteOff :: MutableByteArray# -> Int# -> State# s -> _` helper functions with `byteArrayContents#`, `plusAddr#`, and `*OffAddr#`. For example:

```hs
indexByteOffPrim :: Prim a => ByteArray# -> Int# -> a
indexByteOffPrim arr# idx# =
  case byteArrayContents# arr# of
    addr# -> indexOffAddr# (plusAddr# addr# idx#) 0
```

```hs
writeByteOffPrim :: Prim a => MutableByteArray# -> Int# -> a -> State# s -> State# s
writeByteOffPrim marr# idx# a s0# =
  case unsafeFreezeByteArray# marr# st0# of
    (# s1#, arr# #) -> case byteArrayContents arr# of
      addr# -> writeOffAddr# (plusAddr# arr# idx# a) 0 s1#
```

lehins:
@qrpnxz Your pinned solution would not work for `vector` first of all. Secondly why would use `Primitive` for pinned memory, you can just use `Storable` vector. 

I know for certain that this problem is solvable with primitive operations that work on a byte offset, such as [`indexWord8ArrayAsWord16#`](https://downloads.haskell.org/~ghc/8.10.7/docs/html/libraries/base-4.14.3.0/GHC-Exts.html#v:indexWord8ArrayAsWord16-35-) for example, however this is not a discussion for `vector` package, you should open an issue on the [`primitive`](https://github.com/haskell/primitive) repository if you really care about this.

lehins:
The whole point of [`Data.Vector.Primitive.Vector`](https://hackage.haskell.org/package/vector-0.13.0.0/docs/Data-Vector-Primitive.html#t:Vector) is that it is unpinned memory. If you need pinned memory you should use [`Data.Vector.Storable.Vector`](https://hackage.haskell.org/package/vector-0.13.0.0/docs/Data-Vector-Storable.html#t:Vector) vector.

Switching primitive vector to use byte offset is a serious breaking change, which could happen, but only if primitive supported it and there was enough motivation to get this done.
Switching primitive to use pinned memory in order to achieve zero copy to Text and ByteString does not outweigh the issue of memory fragmentation when a lot of small vectors are being used. This is not something we will do today or ever for that matter!

FYI:
* It is possible to cast Primitive vector into Text without copy, since both use unpinned memory.
* It is possible to cast Storable vector into ByteString without copy, since both use pinned memory.

>  If the only acceptable solution is to go and change primitive, then yes I can go over there and see what can be done

So yes, the only acceptable solution is to go and change `primitive`

qrpnxz:
Goodness, I did not mean to imply changing the current Primitive to pinned! I was talking about having pinned variant; that is, a whole other type. Of course we still want the unpinned one.

> It is possible to cast Primitive vector into Text without copy, since both use unpinned memory.
> It is possible to cast Storable vector into ByteString without copy, since both use pinned memory.

I said so myself? The plus is being able to do both Text and ByteString.

lehins:
> I was talking about having pinned variant; that is, a whole other type.

Maintaining five different modules with almost the same API already takes a lot of work, I'd be dreading adding one more module to that list. 

It feels like we keep talking past each other. I recommend you open a ticket explaining clearly:
* the goals you are trying to achieve.
* the proposed solution:
  * list all the benefits we would gain from this solution. If performance is your goal, some basic benchmarks would be useful to support the claim.
  * list all the drawbacks that you can think of
  * estimate or summarize in words how much work it would take to implement the solution and maintain it. Evaluate if all the effort is worth the goal.
* compare it to alternative solutions that you know of

If all maintainers will agree that it really is something worthwhile implementing an maintaining then we can come up with the plan on how to get it done. This would also allow other from the community to pitch in with ideas. Discussing this on a merged PR does not make much sense and it doesn't have the same visibility.

Issue #355, opened 2021-01-16
Separate streams into a separate package


lehins:
This is a long shot, but I thought I'd ask anyways.

`Data.Vector.Fusion.Stream.Monadic` is a quite useful module and I know of at least two packages that could use it without depending on vector, namely `massiv` and `streamly`. There are a few others I found, but they are less maintained.
Another benefit of splitting it into its own package would force us to write some tests and maybe even benchmarks for the stream functionality itself.

Thoughts?

Bodigrim:
Sounds good.

chessai:
+1, have wanted this before

Shimuuar:
+1

> Another benefit of splitting it into its own package would force us to write some tests and maybe even benchmarks for the stream functionality itself.

We'll also have to review document and possible clean up public API of package instead of treating it as implementation detail of vector.

What is place of Bundle in this? Should it go into new package or remain part of vector proper?

lehins:
I am pleasantly surprise to see everyone supporting this!

> What is place of Bundle in this?

Bundle will have to stay inside of `vector`, because `Bundle` depends on `Vector` class:

```haskell
data Chunk v a = Chunk Int (forall m. (PrimMonad m, Vector v a) => Mutable v (PrimState m) a -> m ())
```
But that's ok, because I don't think there is any desire to make Bundle work with anything else other than vector.

Now that I have your guy's support I'll start working on this, but only after I am done with vector-0.12.2.0 release (which is pretty close, I got quite heads way with backporting functionality last night). 

I would really like to see this split before release of vector-0.13.0.0. A bit of bikeshedding: name for new package? Things that come to mind:

* `vector-stream`
* `stream-fusion`
* ...

Shimuuar:
Do you plan to make it part of 0.13 release? If so I'll it to list (compiling it right now) 

Wrt name is have slight preference for `vector-stream` but don't acre much about it.

Shimuuar:
You do. Sorry, missed that

Bodigrim:
+1 to `vector-stream`

Boarders:
Just a random idea, which may be horrible, but how feasible would it be as a long term goal for `text` and `vector` to share this streaming framework?

Shimuuar:
I think it's doable although stream types are subtly different. `text`'s one is strict in both elements and state, and `vector`'s is lazy. Whether it's something that should be done is another question entirely. And much more difficult to answer

PR #405, opened 2021-08-11
Extract vector stream into its own package

lehins:
As discussed in #355 this PR moves the streaming functionality into its own package.

Unfortunately github diff isn't smart enough,  but the actual git history is preserved in the moved `Monadic.hs` file. It is critical to have two separate commits for this to work, despite that one of them doesn't build.

There are no changes to the actual code, except that `Box` type was moved from `Utils.hs` into the new `vector-stream` package.

There are future plans of adding benchmarks and a test suite into the `vector-stream`, but for now to keep review as easy as possible this PR only moves the code.

### DIFF
--- a/vector-stream/README.md
+++ b/vector-stream/README.md
@@ -1,0 +1,6 @@
+The `vector` package [![Build Status](https://travis-ci.org/haskell/vector.png?branch=master)](https://travis-ci.org/haskell/vector)
+====================
+
+An efficient implementation monadic streams used for fusion in `vector` package
+
+See [`vector-stream` on Hackage](http://hackage.haskell.org/package/vector-stream) for more information.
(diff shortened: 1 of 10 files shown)


Shimuuar (approved):

vector-stream/README.md
```diff
@@ -0,0 +1,6 @@
+The `vector` package [![Build Status](https://travis-ci.org/haskell/vector.png?branch=master)](https://travis-ci.org/haskell/vector)
+====================
+
+An efficient implementation monadic streams used for fusion in `vector` package
```
Bodigrim, line 4:
```suggestion
An efficient implementation of monadic streams used for fusion in `vector` package
```

vector-stream/changelog.md
```diff
@@ -0,0 +1,4 @@
+# Changes in version 0.1.0.0
+
+ * Initial move from `vector`, which will depend on this package starting with
```
Bodigrim, line 3:
```suggestion
 * Initial move from `vector`, which will depend on this package starting from
```

Bodigrim (approved):

Issue #410, opened 2021-08-30
Storable.Mutable.set operation for Vector Double does not work correctly on ghcjs


dfordivam:
With ghcjs the below code results in output `[1.0,5.299808824e-315]`

```
import qualified Data.Vector.Storable as S
import qualified Data.Vector.Storable.Mutable as SM

vecSet = do
  v <- SM.new 2
  SM.set v (1.0 :: Double)
  print =<< S.freeze v
```

The problematic code is here, this causes the use of primitive `Word64` operations which dont work for `Double` as their underlying representation is different.

https://github.com/haskell/vector/blob/f2339ea0c588f6e3c11c563c1cb1e2dcfc316e16/vector/src/Data/Vector/Storable/Mutable.hs#L217

lehins:
Fixed in #411

lehins:
@dfordivam In case you need this fix, I just released a minor fixup version that includes this bugfix: [`vector-0.12.3.1`](https://hackage.haskell.org/package/vector-0.12.3.1)

PR #411, opened 2021-08-30
With ghcjs dont use prim ops of Word64 (for size 8) in storableSet API

dfordivam:
A fix for #410


lehins:
I don't really have a way to validate this change, but it looks reasonable to me. Unless anyone opposes in the next few days, I'll go ahead with merging it.

PR #413, opened 2021-09-08
Use coerce + type application to define vector instances

Shimuuar:
Since we've dropped support for 7.10 we can use type application to pick methods to coerce. I think it result in much clearer code: we drop all wrapping/unwrapping line noise. Hopefully it will result in faster compilation since GHC does not need to inline all fmap and newtype constructors/pattern matching (not tested)

Also it turns out few INLINE pragmas were missing. I wish we had INLINE_ALL pragms

### DIFF
--- a/vector/src/Data/Vector/Unboxed/Base.hs
+++ b/vector/src/Data/Vector/Unboxed/Base.hs
@@ -341,46 +341,52 @@ newtype instance MVector s (As a b) = MV_UnboxAs (MVector s b)
 newtype instance Vector    (As a b) = V_UnboxAs  (Vector b)
 
 instance (IsoUnbox a b, Unbox b) => M.MVector MVector (As a b) where
+  -- Methods that just use underlying vector
   {-# INLINE basicLength #-}
   {-# INLINE basicUnsafeSlice #-}
   {-# INLINE basicOverlaps #-}
   {-# INLINE basicUnsafeNew #-}
   {-# INLINE basicInitialize #-}
+  {-# INLINE basicUnsafeCopy #-}
+  {-# INLINE basicUnsafeMove #-}
+  {-# INLINE basicUnsafeGrow #-}
+  {-# INLINE basicClear #-}
+  basicLength      = coerce $ M.basicLength      @MVector @b
+  basicUnsafeSlice = coerce $ M.basicUnsafeSlice @MVector @b
+  basicOverlaps    = coerce $ M.basicOverlaps    @MVector @b
+  basicUnsafeNew   = coerce $ M.basicUnsafeNew   @MVector @b
+  basicInitialize  = coerce $ M.basicInitialize  @MVector @b
+  basicUnsafeCopy  = coerce $ M.basicUnsafeCopy  @MVector @b
+  basicUnsafeMove  = coerce $ M.basicUnsafeMove  @MVector @b
+  basicUnsafeGrow  = coerce $ M.basicUnsafeGrow  @MVector @b
+  basicClear       = coerce $ M.basicClear       @MVector @b
+  -- Conversion to/from underlying representation
   {-# INLINE basicUnsafeReplicate #-}
   {-# INLINE basicUnsafeRead #-}
   {-# INLINE basicUnsafeWrite #-}
-  {-# INLINE basicClear #-}
   {-# INLINE basicSet #-}
-  {-# INLINE basicUnsafeCopy #-}
-  {-# INLINE basicUnsafeGrow #-}
-  basicLength (MV_UnboxAs v) = M.basicLength v
-  basicUnsafeSlice i n (MV_UnboxAs v) = MV_UnboxAs $ M.basicUnsafeSlice i n v
-  basicOverlaps (MV_UnboxAs v1) (MV_UnboxAs v2) = M.basicOverlaps v1 v2
-  basicUnsafeNew n = MV_UnboxAs `liftM` M.basicUnsafeNew n
-  basicInitialize (MV_UnboxAs v) = M.basicInitialize v
-  basicUnsafeReplicate n (As x) = MV_UnboxAs `liftM` M.basicUnsafeReplicate n (toURepr x)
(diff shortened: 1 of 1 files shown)


vector/src/Data/Vector/Unboxed/Base.hs
```diff
@@ -666,7 +675,7 @@ instance inst_ctxt => G.Vector Vector (inst_head) where { \
 ; basicUnsafeSlice i n (con v)      = con $ G.basicUnsafeSlice i n v      \
 ; basicUnsafeIndexM (con v) i       = tyC `liftM` G.basicUnsafeIndexM v i \
 ; basicUnsafeCopy (mcon mv) (con v) = G.basicUnsafeCopy mv v              \
-; elemseq _ (tyC a)                 = G.elemseq (undefined :: Vector a) a \
+; elemseq _ (tyC a)                 = G.elemseq (undefined :: Vector x) a \
```
lehins, line 678: I can't figure out where is this `x` is coming from? Why is this change necessary?

lehins (approved):
Overall looks great. I only had one question. Also can you rebase on master, so we can get CI green?

vector/src/Data/Vector/Unboxed/Base.hs
Shimuuar: That's because of `ScopedTypeVariables`. If type variable `a` is mentioned in instance head GHC will think that undefined here is of type `Vector a` with `a` same as in head instead of some other `a`. `x` is just way to avoid name capture.

Hopefully we'll be able to get rid of most of the CPP in the future.

lehins: I just merged the PR, but now I am thinking that this `a` was there intentionally. For example instance for `Arg`:

```haskell
instance (Unbox a, Unbox b) => G.Vector Vector (Arg a b) where
...
  elemseq _ (Arg x y) z          = G.elemseq (undefined :: Vector a) x
                                 $ G.elemseq (undefined :: Vector b) y z
```

lehins: I do hope we can get rid of all this CPP nonsense some day.

Shimuuar: This is quite confusing but in the end it's simple. Let take look at `elemseq`'s type:
```haskell
elemseq :: v a -> a -> b -> b
```
Note that type of vector's element could be inferred either from first of second argument. Now let look at macro expansions. For say `All` and `Down a` CPP produces
```haskell
instance G.Vector Vector All where
   elemseq _ (All a) = G.elemseq (undefined :: Vector a) a 
instance G.Vector Vector (Alt f a) where
   elemseq _ (Alt a) = G.elemseq (undefined :: Vector a) a 
```
When `ScopedTypeVariables` is off `a` is fresh type variable in both cases and GHC just unify it and infer correct type. When `ScopedTypeVariables` is on it's still fresh in first case but in second case it's bound in the instance head and it's wrong. 

```haskell
-- Correct
G.elemseq (undefined :: Vector (f a)) (a :: f a)
-- Generated by CPP
G.elemseq (undefined :: Vector a) (a :: f a)
```
For `Arg` annotations just happened to be correct. This is good old case of name capture by macros.

lehins: Yeah, I get the part of scoped variable capture of `a`. What I don't understand where `x` comes from. Is it a type variable or cpp argument, what is it? I don't see it anywhere in scope. What I'd expect is for `newtypeVector` to actually get the type variable as an argument:
```haskell
#define newtypeVector(inst_ctxt,inst_head,rep,tyC,con,mcon) \
instance inst_ctxt => G.Vector Vector (inst_head) where { \
...
; elemseq _ (tyC a)                 = G.elemseq (undefined :: Vector rep) a \
...
#define deriveNewtypeInstances(inst_ctxt,inst_head,rep,tyC,con,mcon) \
...
newtypeVector(inst_ctxt, inst_head, rep, tyC, con, mcon)
```

Shimuuar: I'm not sure myself but GHC must insert quantifier somewhere. Only question is where? After looking at the core I think it's `G.elemseq (undefined :: forall a. Vector a)`. And we naturally can pass such value to function expecting `Vector A`

Let take following simplified example:
```haskell
class Vec v a where
  elemseq :: v a -> a -> b -> b

instance Vec Maybe a where
  elemseq _ = seq

instance Vec [] a where
  elemseq _ a = elemseq (undefined :: Maybe x) a
```

Core for last instance just after desugaring is
```
$celemseq_aCo :: forall a b. [a] -> a -> b -> b
$celemseq_aCo
  = \ (@ a_aCm) (@ b_aCq) ->
      let {
        $dVec_aCu :: Vec Maybe a_aCm
        $dVec_aCu = QWE.$fVecMaybea @ a_aCm } in
      \ (ds_d1pa :: [a_aCm]) (a_aAS :: a_aCm) ->
        break<0>(a_aAS)
        elemseq
          @ Maybe
          @ a_aCm
          $dVec_aCu
          @ b_aCq
          ((\(@ x_aCD) -> undefined @ 'GHC.Types.LiftedRep @ (Maybe x_aCD) CALL_STACK) @ a_aCm)
          a_aAS
```
Note last few line. We do pass type into it. So in original program it must be `undefined :: forall x. Maybe x`

Here is another example:
```haskell
foo xs = (0::Int) + sum (xs :: [x])
```
It fails to compile with 
```
Couldn't match expected type ‘[x]’ with actual type ‘p’
        because type variable ‘x’ would escape its scope
```
I interpret this as "xs is not `forall x. [x]` when it's most certainly `[Int]`"

lehins: I finally get it! GHC will infer the type of `x` from the value `a`:
```
elemseq _ (tyC a)                 = G.elemseq (undefined :: Vector x) a 
```
because of second argument `a` is the same as in `v a`:
```
elemseq :: v a -> a -> b -> b
```

So it seems that the only type annotation we need is `elemseq @Vector undefined`. So, now it makes total sense and I 100% support the change you made. 

Thank you for bearing with me.

PR #414, opened 2021-09-08
Remove dead comments

Shimuuar:
1. Type class is defined in D.V.Generic.Base. No reason to keep its commented out code
2. Remove obsolete comment. When offending function (Data.Vector.Unboxed.unstream . Bundle.fromList) is compiled it produces rather straightforward core. It seems GHC fixed problem whatever it was.
3. Typo in error message


lehins (approved):
I wanted to remove that stuff for the longest time, but kept forgetting about. Thank you for finally getting rid of it.

Some weird issue on CI, might be borked cache. I restarted it, maybe it will fix itself

Shimuuar:
> Some weird issue on CI, might be borked cache. I restarted it, maybe it will fix itself

It didn't. GHC 8.6.5 fails quite reliably on ubuntu with 
```
vector-doctest: /usr/local/.ghcup/ghc/8.6.5/bin/ghc-8.6.5: getPermissions:getFileStatus: does not exist (No such file or directory)
```

lehins:
This is weird: `.ghcup/ghc/8.6.5/bin/ghc-8.6.5`

`ghcup`? I thought we are using PPA for installing ghc

Shimuuar:
> ghcup? I thought we are using PPA for installing ghc

I'm not sure it's ghcup. But I guess it is. PPA does not for work for windows/mac and we don't treat them specially. I'll try to work around cache(?) problem

lehins:
I get a feeling this recent change might have something to do with it:

https://github.com/actions/virtual-environments/pull/4016/files

Shimuuar:
Maybe. Looks like dropping cache did the trick

Issue #406, opened 2021-08-13
Overflow in size hints in bundles


Shimuuar:
Consider following program:

```haskell
import qualified Data.Vector.Unboxed as U

main :: IO ()
main = do
  print $ U.take 4 $ U.generate (maxBound - 4) id
                U.++ U.generate (maxBound - 4) id
```

One could think that it will print `[0,1,2,3]` provided it will fuse.  It however fails with.

```
overflow: Data.Vector.Fusion.Bundle.Size.checkedAdd: overflow: -10
CallStack (from HasCallStack):
  error, called at src/Data/Vector/Fusion/Bundle/Size.hs:56:7 in vector-0.13.0.1-KtkGVThkqZeLj5OVQdyUBP:Data.Vector.Fusion.Bundle.Size
```

This happens because computation of both exact vector length and upper bound on vector size throws error if it overflows `Int`. I wasn't able to craft example which overflows maximum bound on length which is much more realistic. 

This issue is mostly to record that error on error on overflow could cause problems. I'd propose following semantics change: overflow for upper bound on vector size reverts to unbounded. Overflow for lower bound does cause error. (Note not expressible with current implementation of size)

Related: #388

Issue #388, opened 2021-05-23
Allocation strategies for vector creation


Shimuuar:
Currently we have following size hints in bundle:

```haskell
data Size = Exact Int          -- ^ Exact size
          | Max   Int          -- ^ Upper bound on the size
          | Unknown            -- ^ Unknown size
```

however buffer allocation for vector has only two strategies: doubling for unknown and exact  allocation for both `Exact` and `Max`. I think we should have three: unbounded doubling, doubling with bound for Max, and preallocation for Exact

P.S. It was also proposed to add lower bound to  [discussion](https://github.com/haskell/vector/issues/301#issuecomment-845587128)

gksato:
From [your comment in #301](https://github.com/haskell/vector/issues/301#issuecomment-821949039), parhaps you intend to assign `Exact` for `fromListN` ?

Shimuuar:
Yes but I don't think this is a problem. It's hint for size of underlying buffer which could be larger than array.

gksato:
If we add a new field `preAlloc`, it's a big breaking change, but I'm afraid of the possibility that changing the semantic behavior like this may break something somewhere. Also, I'm not sure if it is OK to change `Data.Vector.Generic.length` from O(1) to O(N). Currently we have
``` Haskell
Data.Vector.Fusion.Bundle.length Bundle{ sSize = Exact n } = n
```
What if someone relies on it? I can't imagine there are any; I can explain myself that there couldn't be any, but I can't convince myself.

gksato:
Oh no. We have
``` Haskell
Generic.length = Bundle.length . Generic.stream
```
That is, in order to calculate in constant time the length of an actually heap-allocated vector, we rely on the fact that `Exact` holds the exact length. We need to implement a big detour in order to make that semantical change.

Shimuuar:
> Generic.length = Bundle.length . Generic.stream

You're right. That's bad on one hand but isn't big problem. We just need to make hint more precise. It could be solved by adding one more constructor:
```haskell
...
  | Exact Int -- Vector will have exactly this length
  | Preallocate Int -- Allocate buffer of this size
```
And is we extend size hints with lower bounds on size bounds it could be written as `Max n n`

gksato:
@Shimuuar Correct. I think we agree on the fact that we can't keep the type `Size` as it is. However, there could be many choices we can make; what you proposed in is surely one thing:

``` Haskell
data Size
   = Exact { length :: Int } -- exact size
[...]

PR #418, opened 2021-09-14
Add tests for munstream 

Shimuuar:
While working on unstreaming machinery (See #301, #388, #406 for context) I found out that we don't test `unstream` at all. One could replace its definition with undefined and tests will still pass. This PR fixes this situation by adding tests that `unstream`, `vunstream` and `unstreamR` produce identical results. 

There're few cleanups for the test suite.


lehins (approved):
This looks great. Less CPP makes me happy :smile:

PR #423, opened 2021-11-17
Bump dependency range for doctest (0.19 is released)

Shimuuar:
Sadly it doesn't fixes problems with recent GHC versions. Hopefully 0.20 will


lehins (approved):

lehins:
@Shimuuar I just checked after merging and the `0.20.0` is also already out.

Shimuuar:
Great! I'll try it out. Maybe it did fix our tests

PR #425, opened 2021-12-05
Add safe read variant for mutable vectors

### LINKED ISSUE #341: Safe mutable read variant
It would be nice if `Data.Vector.Mutable` (and variants) had a safe read option (or perhaps read with default) of the form (name is irrelevant to me)

```haskell
safeRead :: PrimMonad m => MVector (PrimState m) a -> Int -> m (Maybe a)
```

Shimuuar:
Also add doctests for read along the way and bump doctest's upper bound (0.21) is on hackage

Fixes #341

### DIFF
--- a/vector/src/Data/Vector/Generic/Mutable.hs
+++ b/vector/src/Data/Vector/Generic/Mutable.hs
@@ -635,7 +635,15 @@ clear = stToPrim . basicClear
 -- Accessing individual elements
 -- -----------------------------
 
--- | Yield the element at the given position.
+-- | Yield the element at the given position. Will throw exception if
+-- index is out of range.
+--
+-- ==== __Examples__
+--
+-- >>> import qualified Data.Vector.Mutable as MV
+-- >>> v <- MV.generate 10 (\x -> x*x)
+-- >>> MV.read v 3
+-- 9
 read :: (HasCallStack, PrimMonad m, MVector v a) => v (PrimState m) a -> Int -> m a
 {-# INLINE read #-}
 read v i = checkIndex Bounds i (length v)
--- a/vector/src/Data/Vector/Generic/Mutable.hs
+++ b/vector/src/Data/Vector/Generic/Mutable.hs
@@ -44,7 +44,7 @@ module Data.Vector.Generic.Mutable (
   clear,
 
   -- * Accessing individual elements
-  read, write, modify, modifyM, swap, exchange,
+  read, readMaybe, write, modify, modifyM, swap, exchange,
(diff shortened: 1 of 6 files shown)


vector/src/Data/Vector/Primitive/Mutable.hs
```diff
@@ -386,11 +386,34 @@ clear = G.clear
...
 
--- | Yield the element at the given position.
+-- | Yield the element at the given position. Will throw exception if
+-- index is out of range.
+--
+-- ==== __Examples__
+--
+-- >>> import qualified Data.Vector.Primitive.Mutable as MVP
+-- >>> v <- MV.generate 10 (\x -> x*x)
+-- >>> MV.read v 3
```
lehins, line 396:
```suggestion
-- >>> v <- MVP.generate 10 (\x -> x*x)
-- >>> MVP.read v 3
```

lehins (approved):
Maybe we should call it `readMaybe`? Just a suggestion.

Other than that (and minor doctest fix) looks great.

lehins:
Oh yeah, one more thing. Adding `since` pragmas would be nice, but not necessary. I can add them when doing a release.

vector/src/Data/Vector/Generic/Mutable.hs
```diff
@@ -635,12 +635,36 @@ clear = stToPrim . basicClear
 -- Accessing individual elements
 -- -----------------------------
 
--- | Yield the element at the given position.
+-- | Yield the element at the given position. Will throw exception if
+-- index is out of range.
```
konsumlamm, line 639:
```suggestion
-- | Yield the element at the given position. Will throw an exception if
-- the index is out of range.
```
Same for the other versions.

vector/src/Data/Vector/Generic/Mutable.hs
```diff
@@ -635,12 +635,36 @@ clear = stToPrim . basicClear
...
+-- >>> v <- MV.generate 10 (\x -> x*x)
+-- >>> MV.read v 3
+-- 9
 read :: (HasCallStack, PrimMonad m, MVector v a) => v (PrimState m) a -> Int -> m a
 {-# INLINE read #-}
 read v i = checkIndex Bounds i (length v)
          $ unsafeRead v i
 
+-- | Yield the element at the given position. Returns 'Nothing' if
+-- index is out of range.
```
konsumlamm, line 653:
```suggestion
-- the index is out of range.
```
Same for the other versions.

Shimuuar:
Thank you! I've applied your suggestions. (And yes `readMaybe` is better name)

Issue #361, opened 2021-01-18
Migrate unsafe functions into their own Unsafe modules


lehins:
This is a proposal and hasn't been set in stone yet. Please feel free to comment your opinion on the matter in this ticket.

Prelude: It has been requested many times that we provide access to all constructors from `vector` package. Giving access to those constructors allows violation of many invariants and hence is unsafe. Quote from #357:

> Decide what is best way to expose constructors for vectors. It have been requested numerous times: #245, #343, #171, #63 (tangentially related), #49. It's of course totally unsafe. Still it's very useful to have on occasion

The best approach to tackle this is to create special `Unsafe` modules and export such constructors form there. This also spawned a discussion that it would make sense to migrate all of the of the unsafe functions to such modules (eg. `unsafeIndex`, `unsafeRead`, etc.), in order to provide a clear boundary between safe and dangerous functionality, which can have different qualified imports in the user land.

Doing such migration of unsafe functionality is worthwhile in a long run, but immediate problem is that it will cause a lot of breakage. Therefore it would have to be done in a few stages across couple of major releases.

1. Next minor release `vector-0.12.2.0` would be adding these modules with all of the unsafe functions exported from there:
```haskell
module Data.Vector.Unsafe
module Data.Vector.Generic.Unsafe
module Data.Vector.Primitive.Unsafe
module Data.Vector.Storable.Unsafe
module Data.Vector.Unboxed.Unsafe
```
2. Next major release `vector-0.13`, would continue to export all unsafe function from the Unsafe modules and their current locations, but latter ones would get `DEPRECATED` pragma added to them. This way such change will be visible to all the users
3. The following major release `vector-0.14` would have those unsafe functions removed from their current locations, but only if such release happens at least a year after `vector-0.13` that deprecated unsafe functions.

Other relevant points to discuss:
1. Do we stretch out the deprecation timeline over a longer period of time (eg. announce in vector-0.13, deprecated in vector-0.14 and remove in vector-0.15) It will take a couple of years.
1. Do we also provide Unsafe modules for Bundle?
1. Do we have separate Unsafe modules for Mutable vs immutable. My opinion is that such separation is not needed, since there aren't any name clashes of top of my head. And if there are any they can be mitigated by renaming, since this change is additive.
1. Should we bother with monomorphic unsafe exports or instead only export `Generic` unsafe functions?
1. Is the whole thing really worth it?

Shimuuar:
[...]

Issue #357, opened 2021-01-17
Release plan for 0.13


Shimuuar:
Here is list of issues that need to be fixed before 0.13 release. I only list breaking changes and omit things that could be added to 0.13.1, etc.

 - [x] Split streams into separate package #355
 - [x] Decide what is best way to expose constructors for vectors. It have been requested numerous times: #245, #343, #171, #63 (tangentially related), #49. It's of course totally unsafe. Still it's very useful to have on occasion
 - [x]  Determine semantics of slice/grow/etc #276, #36. Likely it's going to be a documentation change but it's better to do this for 0.13 just in case.
 - [x] unfoldrN, unfoldrNM and fromListN are dangerous #301. Could require change to semantics of functions. 
 - [x]  The Bundle datatype seemingly doesn't need the field sVector #269. Not terribly important but certainly breaking
 - [x] decide semantics of maximumBy & friends

lehins:
We'll hold off on #269 it might be useful to keep `sVector` afterall.

lehins:
I'd like to communicate who's going to be working on which tickets so we don't start stomping on each other's feet.

I'll can tackle the #36 now, since I recently had to deal with those semantics.

I already mentioned it on a ticket, but I'll be happy to do the split described in #355 

With regards to #276 I could take care of it too, since I was involved in that discussion and also fixed the original issue with slice in #257, **however** it would be great to get another pair of eye on the issue, so maybe someone else can pick it out.

#306 I suggested a solution: https://github.com/haskell/vector/issues/301#issuecomment-592279595 which is the approach I've taken in massiv, so if someone would like to implement and document it, that'd be great.

With regards to exporting constructors, I'd go for adding these modules:
```haskell
module Data.Vector.Internal
module Data.Vector.Primitive.Internal
module Data.Vector.Storable.Internal
module Data.Vector.Unboxed.Internal
```
that have `{-# OPTIONS_HADDOCK hide, not-home #-}`

An easy way would be to simply re-export constructors from there or the hard way of moving constructors with all the instances into those module in order to avoid orphans. Former seems a bit less intrusive. This one is also up for the taking.

lehins:
One a second thought it might be a good idea to move all of the unsafe functions into their own modules, which would also include constructors.

```module
module Data.Vector.Unsafe
module Data.Vector.Primitive.Unsafe
module Data.Vector.Storable.Unsafe
module Data.Vector.Unboxed.Unsafe
```

[...]

Issue #245, opened 2019-05-15
Export data constructor for boxed vector


andrewthad:
The module `Data.Vector` does not export the data constructor for `Vector`. Exporting this would make it possible to use `vector`'s functions in contexts where I am working with an unsliced `Array` and want to fold over a slice of it. In general, I would like to see all of the data constructors exported.

@cartazio In https://github.com/haskell/vector/issues/49, this is the same position you take. Do you still agree?

chessai:
I have also wanted this on numerous occasions.

cartazio:
Ill reread everything and then think about it. Don’t have vector loaded
into my head atm.

But speculative yes to making the constructor public at least in an
internal module

Shimuuar:
I think exporting constructor from Data.Vector is wrong. It's trivial to create memory unsafe vector value: `Vector 0 10000 []`. It's probably better to export it from some `*.Internal` module. It would make obvious that it is dangerous.

andrewthad:
I don't care what module it comes from. Either `Data.Vector.Unsafe` or `Data.Vector.Internal` would be fine to me.

cartazio:
Yeah. We’re all on the same page.

lehins:
Fixed in #434 by providing functions that give access to boxed Vector's internal `Array`

Issue #49, opened 2014-10-16
Vector.Unboxed doesn't export constructors, but Unboxed.Mutable does


cartazio:
I think they should both be exported (or at least I dont see any good reason not to export them)

lehins:
Constructors for unboxed vectors have been exported from `Data.Vector.Unboxed.Base` from the beginning of time. It is an internal module as it very much should be, but it does not prevent others to shoot themselves in the foot.

PR #434, opened 2022-05-22
Add `unsafeToArray` function for getting the underlying boxed `Array`

lehins:
#245 requested access to `Vector` constructor. There are a few other requests like that, see this https://github.com/haskell/vector/issues/357#issue-787733102

However, there is a universal agreement that ability to construct a very unsafe Vector with a regular type constructor is unsettling to say the least. See this: https://github.com/haskell/vector/issues/245#issuecomment-492753908 and this https://github.com/haskell/vector/issues/49#issuecomment-62709758 Let's be honest, creating an `Internal` module just for constructors is not going to happen and separating all unsafe functions into `Unsafe` module is lost cause too, see #361

The next best thing we can do, I think, is to create a couple of `unsafe*` functions that let users create/access underlying buffers in the unsafe way, similar to `unsafeToForeignPtr`.

This PR implements what I suggest, but only for immutable boxed vector. Mutable vector constructor have been exported since inception, so might as well leave it be.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,28 +18,28 @@ jobs:
       matrix:
         include:
           # Linux
-          - { cabal: "3.4", os: ubuntu-latest,  ghc: "8.0.2"  }
-          - { cabal: "3.4", os: ubuntu-latest,  ghc: "8.2.2"  }
-          - { cabal: "3.4", os: ubuntu-latest,  ghc: "8.4.4"  }
-          - { cabal: "3.4", os: ubuntu-latest,  ghc: "8.6.5"  }
-          - { cabal: "3.4", os: ubuntu-latest,  ghc: "8.8.4"  }
-          - { cabal: "3.4", os: ubuntu-latest,  ghc: "8.10.7" }
-          - { cabal: "3.4", os: ubuntu-latest,  ghc: "9.0.1"  }
-          - { cabal: "3.4", os: ubuntu-latest,  ghc: "9.2.1"  }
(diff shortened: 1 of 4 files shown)


chessai:
This looks good, but I've always wanted the following in vector as well:

```haskell
-- performs a copy, for when I no longer need slicing
toArray :: Vector a -> Array a

-- provided by this PR, not sure what warrants "unsafe". surely using the result of this function would be potentially unsafe, but this function alone shouldn't be, as far as I can tell
unsafeToArray :: Vector a -> (Array a, Int, Int)

-- construct a vector with offset 0 and length equal to the array
fromArray :: Array a -> Vector a

-- unsafe because of manual slicing
unsafeFromArray :: Array a -> Int -> Int -> Vector a
```

Shimuuar:
Indeed, there's nothing unsafe with `unsafeToArray`. Only construction is unsafe. And we may add constructor in both unsafe and variety

```
-- Throws exception if slice is invalid
fromArraySlice :: Array a -> Int -> Int -> Vector a
-- unsafe because of manual slicing
unsafeFromArraySlice :: Array a -> Int -> Int -> Vector a

```

lehins:
Addressed all the feedback. This PR is ready for a followup review.

Shimuuar (approved):
LGTM

Another option is add read-only pattern like one below. It allows to look at vector internals but doesn't allow construction.

```haskell
pattern Vector :: forall a. Int -> Int -> Array a -> Vector a
pattern Vector off sz arr <- MkVector off sz arr
{-# COMPLETE Vec #-}
```

Issue #370, opened 2021-03-05
Fix doctests on GHC 8.10 and GHC 9.0


Shimuuar:
At least on CI they don't work at all and fail with something like that:

```
Data/Vector/Storable.hs:558: failure in expression `import qualified Data.Vector.Storable as VS'
expected: 
 but got: 
          ^
          <interactive>:1:1: error:
              attempting to use module ‘main:Data.Vector.Storable’ (Data/Vector/Storable.hs) which is not loaded
```

Cursory search doctest's issue tracker does reveal anything interesting

Shimuuar:
I tried to run doctests on 8.10 locally. There're weird error when something is not in scope when it was imported  line above. On top of that they become very slow. First 141 test pass very quickly after that they slow to a crawl

Situation with doctests certainly doesn't look good.

Bodigrim:
I lost my hope in doctests, they are so difficult to setup properly on recent GHCs :(

Shimuuar:
They probably couldn't work properly without explicit support from GHC

lehins:
I just disabled doctests in `random` for newer ghc: https://github.com/haskell/random/blob/b89d4117797e646ab55a8fece5b468a2d5b10682/test/doctests.hs#L4-L10

So, I agree situation with doctests is not great. However, I think it is important to run doctests at least on one ghc version. The point of doctests is not to test the underlying representation, that's what unit, property ... tests are for, but to make sure that examples in documentation are still correct.

I suggest we do the same as I did in `random`. Disable doctests on newer ghc versions, until situation improves. Maybe that could be one of those things that Haskell Foundation could help find people interested in figuring out a solution.

Shimuuar:
That's unfortunate status quo. They're disabled for 8.10 & 9.0. Also I completely agree that we do need doctests

Shimuuar:
Fixed by #465

PR #435, opened 2022-06-12
Fix doctests for GHC 8.10 and 9.0

HanStolpo:
Fix doctests to work for later versions of GHC. The cabal file specifies
the default language to be Haskell2010 but doctest seems to not pickup
this configuration. We fix this by passing this in as a command line
argument to doctest.

Relates to github issue https://github.com/haskell/vector/issues/370


lehins:
Awesome!!!! Thank you Handre

PR #462, opened 2023-06-04
Future-proof against potential Prelude.foldl'

Bodigrim:
See https://github.com/haskell/core-libraries-committee/issues/167

This is, of course, a bit speculative at the moment, but given that the change does not involve CPP, it should not hurt to merge it early to simplify further impact assessment.


Shimuuar:
LGTM but build for GHC<=8.6 fails due to fail not being in scope

Bodigrim:
@Shimuuar CI is green.

PR #463, opened 2023-07-27
Allow deepseq-1.5

Bodigrim:
It's a bit challenging to test at the moment: I had to build against a patched version of `primitive`, which does not depend on `template-haskell`.


lehins (approved):
:+1:

PR #465, opened 2023-08-27
Optimize replicateM and update CI

Shimuuar:
This PR turn out to be a bit of kitchen sink

1. Use existing `munstream` instead of hand-rolled loop in `replicateM` for mutable vectors. It also avoid double initialization of buffer with `new` (noticed by @dmalkr) 
2. Small cleanups in test suite: add default value for `TestEq` and default implementation for `equals` method of `TestData`  type class. 
3. Then I noticed that we only test GHC up to 9.2 on CI. I've update CI config


Shimuuar:
And look at it! Doctests blew up again. I'll look into it tomorrow

Shimuuar:
I managed to fix doctests on GHC 9.4 & 9.6. It's a big hack but it does work. It seems there's state leakage between running tests on different modules and it caused tests to fail

lehins (approved):
Great find about on doctests.

LGTM!

PR #469, opened 2023-09-19
Export `findIndexR` from the non-generic vector modules

toyboot4e:
`findIndexR` was only in the generic module and could be missed.

This PR adds type-restricted versions of  `findIndexR` to the non-generic vector modules (plain `Vector`, `Primitive`, `Storable` and `Unboxed`). It also adds `Does not fuse.` sentences to their documents.

Related: #172


Shimuuar:
Thank you!

PR #473, opened 2023-12-01
Add instances for Mutable t.f. for arrays from primitive

Shimuuar:
While arrays from primitive could not have `Vector`/`MVector` instances, `Mutable` is well defined for them. It's useful on its own and could be used with another pair of type classes.

### DIFF
--- a/vector/src/Data/Vector/Generic/Base.hs
+++ b/vector/src/Data/Vector/Generic/Base.hs
@@ -28,6 +28,9 @@ module Data.Vector.Generic.Base (
 import           Data.Vector.Generic.Mutable.Base ( MVector )
 import qualified Data.Vector.Generic.Mutable.Base as M
 import           Data.Vector.Fusion.Util (Box(..), liftBox)
+import qualified Data.Primitive.Array as Prim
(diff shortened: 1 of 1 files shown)


Shimuuar:
@lehins, @Bodigrim  Any objections?

lehins:
To be honest with you, I fail to see the usefulness of such type family instance for primitive types, but at the same time, I don't see anything wrong with it either.
It should work for all primitive arrays, except for ByteArray, since the kind does not match for it.

lehins:
Maybe we should extract `Vector v => Slicable v` into a separate class?

Bodigrim:
> To be honest with you, I fail to see the usefulness of such type family instance for primitive types, but at the same time, I don't see anything wrong with it either.

+1.

Shimuuar:
> Maybe we should extract slicing functionality `Vector v => Slicable v` into a separate class?

It's something that I would like to do. But it's rather complicated, it's quite invasive change which will break all unboxed instances. There're functions like `takeWhile` which could be implemented in stream fusion framework and as a slice. API for creating slices is quite anemic too. I really should make a writeup about this

PR #476, opened 2024-01-16
Implement spanR & breakR

### LINKED ISSUE #172: Is there an efficient way to `break`/`span`/`findIndex` from the right end?

Shimuuar:
They follow same convention as `Data.Sequence.spanR/breakR`. Fixes #172

I also notices that we don't run tests with unsafe and internal range checks on. We should do that in part because we should check that such configuration actually works and it may save us in the future.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,8 @@ jobs:
           - { cabal: "3.10", os: ubuntu-latest,  ghc: "9.2.8"  }
           - { cabal: "3.10", os: ubuntu-latest,  ghc: "9.4.8"  }
           - { cabal: "3.10", os: ubuntu-latest,  ghc: "9.6.3"  }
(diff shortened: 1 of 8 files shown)


lehins (approved):
:+1:

PR #478, opened 2024-01-22
Attempt to run doctests with GHC that were disabled

Shimuuar:
If 624c0ab61b3b69e83c1bc9e58c0003a3e886aea5 did fix doctest with old GHCs we can remove conditionals.


lehins (approved):
Nice!

Shimuuar:
And it blew up on macos. (Linker errors when running doctests)

Issue #459, opened 2023-04-16
Split generic tests into separate package


Shimuuar:
We have an impressive battery of tests which are generic in vector type. I propose splitting them into separate package to benefit anyone who defines instances for vector type classes (me). It turns out creating nontrivial instance *is* error prone. I was writing storable vectors with stride (vector elements are every n-th in buffer)

This will entail creating another package `vector-test` which exposes generic tests and moving vector's test suite there. @lehins @Bodigrim what's you opinion on that?

Bodigrim:
There are https://hackage.haskell.org/package/quickcheck-classes-0.6.5.0/docs/Test-QuickCheck-Classes.html#v:muvectorLaws, it should be easy to make them more polymorphic to cover `Storable`.

Cf. https://github.com/Bodigrim/rotcev/blob/master/test/Main.hs

lehins:
The idea with `quickcheck-classes` would be problematic:
* The package includes tests for laws, where generic test contains all sorts of good reusable tests that aren't related to any laws. This is mostly conceptional problem and might be not that big of a deal.
* `quickcheck-classes` depends on `vector`, so if we move reusable tests into that package, `vector` itself cannot use it, unless we separate `vector` tests into its own package. So, we aren't really winning much, still has to create a `vector-test` package.

I would suggest to create a public sublibrary with reusable tests directly in `vector`. That comes with nice benefits of having all tests stuff still in `vector` without a need to start releasing a new package. The downside we'd have to drop support for building  `cabal < 3.0`. As far I've heard hackage has support for sublibraries nowadays, but I haven't confirmed it personally.

Either way, I do see it as a good idea to to expose all the test functionality that could be useful to the outside world and making it reusable through a library, be it standalone or sublibrary.

Shimuuar:
One benefit of vector's test suite is that it's rather comprehensive. So one could hope that does exercise all relevant properties. It's also more likely to be correct. For example `quickcheck-classes` do not test [tricky part of move](https://hackage.haskell.org/package/quickcheck-classes-0.6.5.0/docs/src/Test.QuickCheck.Classes.MVector.html#writeMoveAroundRead). Namely overlapping source/destination. 

I like idea of using public sublibrary. It however may break some of existing tooling. Quick search brings up: https://github.com/NixOS/cabal2nix/issues/448 https://github.com/NixOS/cabal2nix/issues/539

lehins:
>  It however may break some of existing tooling.

[...]

PR #426, opened 2021-12-05
Benchmarks based on CPU counters

Shimuuar:
We don't run benchmarks with any sort of regularity. In part because it's difficult to get repeatable result, it require dedicated PC and all measurement must be done on same hardware.  So tracking performance regressions is hard. I run into this problem while working on size hints. Another way is to use CPU counters to count number of instructions. It's fast and hopefully reproducible and could possibly run on CI.

This PR is prototype. It sort of works, very rough on edges and right now isn't very deterministic. Apparently GC could kick in and skew results. I need to study this more and to figure out how to ensure repeatable measurements. 

If everything works out I want to split it into separate package so far it's more convenient to develop it as part of vector. 

P.S. `-O1` could be up to 3× slower than `-O2` in our benchmarks but sometimes it's on par.

PR #481, opened 2024-01-28
Expose tests and benchmarks as public sublibrary 

Shimuuar:
This was discussed in #426 and #459 as a way for other packages to access tests and benchmarks from vector. I tested that it's possible to build PAPI-based benchmarks as separate package, but this change is big enough that it deserves its own PR

There're two tests sublibraries: one built with with `-O2` (`vector:tests-O2`) and another with `-O0` (`vector:tests-O0`). In the same way benchmarks library is called `vector:benchmarks-O2`. I want to leave name for benchmarks built with `-O1` since testing both `-O1` and `-O2` could be quite valuable.

Most notable change: building vector requires cabal 3.0 since it's earliest version supporting public sublibraries.

There's also question of naming  of modules in both sublibraries. I left them as they were but it was assumed that they are not available outside of test-suite/benchmarks which is not the case anymore.


Shimuuar:
There's also problem with cabal check. It complains:
```
Warning: These warnings may cause trouble when distributing the package:
Warning: 'ghc-options: -O0' is not needed. Use the --disable-optimization
configure flag.
Warning: 'ghc-options: -O0' is not needed. Use the --disable-optimization
configure flag.
```
and fails CI. It seems there's no good way around this. We do need to build tests with different optimization levels

lehins:
It also complains about bounds:
```
Warning: These packages miss upper bounds:
- base-orphans
- random
- tasty
- tasty-hunit
- tasty-quickcheck
- template-haskell
- transformers
```
I think we can just disable `cabal check`, since those warnings are useless.

Shimuuar:
Cabal check has warnings and warning-but-actually-errors. Former are printed but cabal check returns 0, latter causes non-zero exit code. AFIAR hackage runs cabal check on upload and will reject packages failing it, warnings are accepted. 

Maybe there's way around this by structuring tests different;y. But I think it would be better to keep tests private in this PR

Shimuuar:
Yes I think sublibrary for tests turned out to be trickier than I thought. It would be better to keep tests as they are. And extract common functionality piece by piece later.

With benchmarks situation with optimization flags is more tricky. They export monomorphic function and optimization level is decided when function is built. Application cannot change it.

Shimuuar:
So I left tests mostly alone and only factored shared parts into common stanza. Also:
- Dropped `HUnit` dependency

For benchmarks sublibrary I moved all modules into `Bench.Vector` hierarcy and moved `RandomSeed` & `VectorSize` options to library as well. They will be used by PAPI benchmark as well

lehins (approved):
This looks great!

PR #485, opened 2024-03-11
Hint to GHC that indices are to be used strictly

Bodigrim:
`basicUnsafe{Read/Write/IndexM}` are class members and, unless a call site was already specialised to a specific vector instance, GHC has no clue that the index is most certainly to be used eagerly. Bang before the index provides this vital for optimizer information.

I've been recently looking at Core for programs, where vectors did not specialise to a specific instance for one reason or another. It's really horrible: even if a specific instance remains unknown, we should tell GHC that passing boxed indices around is not OK.


Shimuuar (approved):
Great find! I think `slice`/`unsafeSlice` would benefit from the same treatment as well. 

P.S. Out of curiosity I looking into GHC's timing information. This change reduced allocation by simplifier by about 1e-3 when compiling unboxed vectors.

vector/src/Data/Vector/Generic/Mutable.hs
```diff
@@ -700,21 +700,25 @@ exchange v i x = checkIndex Bounds i (length v) $ unsafeExchange v i x
 -- | Yield the element at the given position. No bounds checks are performed.
 unsafeRead :: (PrimMonad m, MVector v a) => v (PrimState m) a -> Int -> m a
 {-# INLINE unsafeRead #-}
-unsafeRead v i = checkIndex Unsafe i (length v)
+unsafeRead v !i = checkIndex Unsafe i (length v)
                $ stToPrim
```
vaibhavsagar, line 704: I 100% agree with this change, but unfortunately it means the following lines are no longer indented correctly with respect to the `=`. Is that something that can be fixed as part of this PR?

vector/src/Data/Vector/Generic/Mutable.hs
```diff
@@ -700,21 +700,25 @@ exchange v i x = checkIndex Bounds i (length v) $ unsafeExchange v i x
...
+-- The reason is that 'basicUnsafeRead' is a class member and, unless 'unsafeRead' was
+-- already specialised to a specific v, GHC has no clue that i is most certainly
+-- to be used eagerly. Bang before i hints this vital for optimizer information.
 
 -- | Replace the element at the given position. No bounds checks are performed.
 unsafeWrite :: (PrimMonad m, MVector v a) => v (PrimState m) a -> Int -> a -> m ()
 {-# INLINE unsafeWrite #-}
-unsafeWrite v i x = checkIndex Unsafe i (length v)
+unsafeWrite v !i x = checkIndex Unsafe i (length v)
                   $ stToPrim
```
vaibhavsagar, line 715: Same comment as above.

vector/src/Data/Vector/Generic/Mutable.hs
```diff
@@ -700,21 +700,25 @@ exchange v i x = checkIndex Bounds i (length v) $ unsafeExchange v i x
...
                   $ basicUnsafeWrite v i x
 
 -- | Modify the element at the given position. No bounds checks are performed.
 unsafeModify :: (PrimMonad m, MVector v a) => v (PrimState m) a -> (a -> a) -> Int -> m ()
 {-# INLINE unsafeModify #-}
-unsafeModify v f i = checkIndex Unsafe i (length v)
+unsafeModify v f !i = checkIndex Unsafe i (length v)
                    $ stToPrim
```
vaibhavsagar, line 722: Same comment as above.

vector/src/Data/Vector/Generic/Mutable.hs
```diff
@@ -725,7 +729,7 @@ unsafeModify v f i = checkIndex Unsafe i (length v)
 -- @since 0.12.3.0
 unsafeModifyM :: (PrimMonad m, MVector v a) => v (PrimState m) a -> (a -> m a) -> Int -> m ()
 {-# INLINE unsafeModifyM #-}
-unsafeModifyM v f i = checkIndex Unsafe i (length v)
+unsafeModifyM v f !i = checkIndex Unsafe i (length v)
                     $ stToPrim . basicUnsafeWrite v i =<< f =<< stToPrim (basicUnsafeRead v i)
```
vaibhavsagar, line 733: Same comment as above.

Shimuuar:
@Bodigrim I extended same treatment to `unsafeSlice` and converted comment to GHC note. This seems to be a good format for long form comment which are referenced from multiple places in source code

If you're fine with this I'd like to merge this PR

Bodigrim:
@Shimuuar sure, go ahead.

Issue #380, opened 2021-04-16
Implementing a value-strict Vector


infinity0:
Hi, I am trying to implement a variant of `Data.Vector` that is automatically strict in its values.

Since the underlying structure is a primitive `Array#` I cannot add `!` strictness annotations to the data constructors, which would otherwise have been the most efficient way to do this. So I am forced to force every element as it is added to the vector.

It seems that `Generic.Vector` offers a method `elemseq` for this purpose - for `Vector.Storable` this is `seq` whilst for the normal lazy `Vector` this does not evaluate the first subject. So it would seem that to implement a value-strict `Vector` clone, I could simply copy `Data.Vector` with one single addition of definition `elemseq _ = seq`, is that correct? Sorry if this is blindingly obvious but it's not explicitly mentioned in the docs (I suggest to add this & can file a PR to this effect.)

Shimuuar:
No `elemseq` is barely used. You'll have to force element to WHNF in `Vector`/`MVector` instance.

P.S. `Storable` and `Prim` definitions are also not quite correct. Storing value in array will evaluate it to NF, while `seq` evaluates to WHNF. This is more serious for `Storable` where value could mirror some C struct and have lazy fields and that difference would matter.

infinity0:
OK, thanks for the hint. I went through the instance methods and it seems that `MVector`'s `basicUnsafeReplicate` and `basicUnsafeWrite` are the only ones that take in an element to be added to an array, so I just added bangs to those. Is that right? That is still quite simple, I'm happy I didn't have to duplicate the entire top-level API and add bangs everywhere.

gksato:
@infinity0 I remember there is `basicSet`?

infinity0:
@gksato The default implementation is in terms of `basicUnsafeWrite` so I can just rely on that. (I suppose that relies on the default implementation not changing, but there's nothing else it could be defined in terms of.)

gksato:
@infinity0
By setting
```
basicSet (StrictVector v) !x = basicSet v x
```
we can let GHC believe that the caller of `set` may evaluate `x` to WHNF no matter whether `v` is inhabited or empty; that could lead to a better optimization.

infinity0:
I can see what you mean, but what real use-case would that help? IMO it is sufficient that a "strict container" ensures the values *inside* it are forced; this is sufficient to prevent space-leaks. The additional behaviour you're describing doesn't really do anything to prevent space leaks: either

- `x` is lost, thus GCd, or
[...]

PR #488, opened 2024-03-18
Add strict vector

### LINKED ISSUE #483: Add strict boxed vectors
This is inspired by thread on [discource](https://discourse.haskell.org/t/optimiser-performance-problems/8906/22). We don 't have any good way to ensure that elements of boxed vector are evaluated. Namely any use of `map`, `zip`, etc will not evaluate vector's elements. And if used in iterative algorithms as in thread above it will leak memory. We don't even have standard way of evaluating all elements of vector to WHNF.

From users' PoV easiest way is to add boxed vector which is strict in its elements. It would be also most reliable. And since strictness is tied to strictness of `basicWrite` we have to define new data type.

As much as I hate adding (and supporting) another vector data type I think this addition is valuable enough. @lehins @Bodigrim what is your opinion?

I

Shimuuar:
Fixes #483 and #380 

Implementation follows discussion in issues above and virtually identical to one in `strict-containers`. Only difference `fromArray` and `unsafeFromArraySlice` now force all elements of returned vector to WHNF. It changes complexity to O(n) so to compensate `lazyFromArray` and `unsafeLazyFromArraySlice` are added, they retain O(1) complexity but user has to deal with strictness. 

This is draft since there's still some work to do with documentation. 
1. We need to decide to prefix to use in examples. `V.` is used by lazy vector, `VS` by storable. One option
2. I think it would be better to switch examples in generic vector to strict vector. It's better default data structure
3. It's worth making a new release soon after this PR is merged, so I think I should just slap `@since 0.13.2.0` on each function in `Strict` modules

Also
 - Missing LICENSE file for vector-bench-papi
 - Small cleanup in test suite


Shimuuar:
At this point PR is pretty much ready

Bodigrim:
@infinity0 could you possibly review?

Shimuuar:
Basically it's same as version from `strict-containers`. Only differences are strict `fromArray` & `fromMutableArray`, added `lazyFromArray` and haddocks.

I also switched examples in D.V.Generic to strict vectors but that's purely documentation change

vector/src/Data/Vector/Strict/Mutable.hs
```diff
@@ -0,0 +1,828 @@
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE DeriveDataTypeable #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE RoleAnnotations #-}
+{-# LANGUAGE TypeFamilies #-}
+-- |
+-- Module      : Data.Vector.Mutable
```
Bodigrim, line 9:
```suggestion
-- Module      : Data.Vector.Strict.Mutable
```

vector/src/Data/Vector/Strict.hs
```diff
@@ -0,0 +1,2675 @@
...
+
+import qualified Control.Applicative as Applicative
+import qualified Data.Foldable as Foldable
+import qualified Data.Traversable as Traversable
+
+import qualified GHC.Exts as Exts (IsList(..))
+
+
+-- | Boxed vectors, supporting efficient slicing.
+data Vector a = Vector {-# UNPACK #-} !Int
```
Bodigrim, line 215: Could it be a newtype over lazy boxed vectors? I imagine it would save us some instance definitions.

vector/src/Data/Vector/Strict/Mutable.hs
```diff
@@ -0,0 +1,828 @@
...
+  , (+), (-), (*), (<), (>), (>=), (&&), (||), ($), (>>), (<$>) )
+
+import Data.Typeable ( Typeable )
+
+#include "vector.h"
+
+type role MVector nominal representational
+
+-- | Mutable boxed vectors keyed on the monad they live in ('IO' or @'ST' s@).
+data MVector s a = MVector { _offset :: {-# UNPACK #-} !Int
```
Bodigrim, line 93: Same question: could this be a newtype over lazy boxed `MVector`?

Bodigrim (commented):
Awesome!

vector/src/Data/Vector/Strict.hs
Shimuuar: That should be possible. Probably most importantly for cheap conversion between strict and lazy vectors. 

But we should be careful with instances. Anything that produces vectors should be implemented in terms of operations over strict vectors.

Bodigrim: I think it's worth to employ newtypes if possible, to demonstrate clearly that it is the same data structure just with different instances.

Shimuuar:
@Bodigrim switching to a newtype worked out quite well and allowed to drop a lot of duplicated code. It doesn't work all that well for instances since DerivingStrategies are only available since 8.2. So it's not possible to instruct GHC to derive say `Ord`.

I also added function for conversion between strict and lazy vectors.

vector/src/Data/Vector/Strict.hs
```diff
@@ -0,0 +1,2635 @@
...
+--
+-- Maintainer  : Haskell Libraries Team <libraries@haskell.org>
+-- Stability   : experimental
+-- Portability : non-portable
+--
+-- Immutable strict boxed vectors (that is, polymorphic arrays capable
+-- of holding any Haskell value). It is possible to create vector
+-- which contain bottom elements, either by using mutable interfaces
+-- (see "Data.Vector.Strict.Mutable") or functions that don't preserve
+-- strictness ('lazyFromArray').
```
infinity0, line 27: What's the use case for this? Just seems like a potential source of bugs and/or making existing bugs harder to find. If somebody sees this strict type they are going to assume it doesn't contain bottom elements, but this design seems to suggest otherwise.

IMO a type that is advertised as "strict" should have no possible external (user) code path that can lead it to containing non-WHNF elements, even if it's internally implemented using lazy structures.

infinity0: > they retain O(1) complexity but user has to deal with strictness.

I don't think this is a good use-case, if the user selects a strict data type they should be aware that there will be potential performance penalties for certain operations, we shouldn't expose a non-strict operation for them to "regain performance", they should instead just pick the lazy type.

Shimuuar: On one hand it's one of many unsafe functions where all responsibility on maintaining invariants is passed to the caller. So it seems that it would be in a good company.

However here invariant is "all elements are in WHNF" And it's known to be rather tricky to maintain. And even worse it's difficult to check and invariant violations could be invisible when all test look fine and then it fail on prod due to space leak.

So I think it would be better to drop these functions and if someone really need these it could be discussed.

Shimuuar:
I removed function that don't enforce strictness: `lazyFromLazy`, etc.

vector/src/Data/Vector/Strict.hs
```diff
@@ -23,11 +23,9 @@
 -- Immutable strict boxed vectors (that is, polymorphic arrays capable
 -- of holding any Haskell value). It is possible to create vector
 -- which contain bottom elements, either by using mutable interfaces
```
infinity0, line 25: Can drop "either" here now, also might be good to clarify the vectors that might contain bottom elements are the original data type, not this strict type.

Shimuuar: I reworded module comment in order to make it more clear

Shimuuar:
I'm going to merge this PR on next weekend if there are no objections

lehins (approved):
I skimmed through the PR. Looks like it is just more of the same as regular lazy vector. I looked at all the entry points where lazy values could enter into the strict vector and it looks like they are all covered.

So, no objections from me.

PR #489, opened 2024-03-23
Introduce basicUnsafeIndexM#

Bodigrim:
#485 did only half of the job: while GHC now knows that index is used strictly it still would not necessarily unpack it, because `basicUnsafeIndexM` must receive `Int` not `Int#`. Only after inlining an opportunity to erase boxing will arise. 

This patch introduces `basicUnsafeIndexM#` to help GHC further. If it looks good, I'll go for `basicUnsafeRead#` / `basicUnsafeWrite#` in another PR.


Shimuuar:
This is however very breaking change. Due to design of unboxed vector (which I consider a mistake) this change will break all unboxed vectors. They require defining data instances and `MVector/Vector` instances for each type.

Bodigrim:
@Shimuuar what exactly would break? I tried couple of packages with manual instances for unboxed vectors, they seem to compile fine.

Shimuuar:
Sorry, I missed mutually recursive `basicUnsafeIndexM` and `basicUnsafeIndexM#`. I withdraw objection

vector/src/Data/Vector/Fusion/Bundle/Monadic.hs
```diff
@@ -1091,9 +1092,10 @@ fromVector v = v `seq` n `seq` Bundle (Stream step 0)
     n = basicLength v
 
     {-# INLINE step #-}
-    step i | i >= n = return Done
-           | otherwise = case basicUnsafeIndexM v i of
-                           Box x -> return $ Yield x (i+1)
+    step (I# i)
+      | I# i >= n = return Done
```
sergv, line 1096: Here and everywhere in this PR, when we pattern-match on `I#` but then construct it again in function's body is GHC 100% guaranteed to optimise allocation of fresh `Int` away and will just reuse whatever we pattern-matched on?

Shimuuar: GHC _should_ be quite good at this. So it should eliminate `Int` allocations.

Shimuuar:
I tried to measure impact of this optimization. To do so I added simple benchmark which computed variance of vector of doubles (branches `index-profiling` and `index-profiling-B`) in two otherwise identical variants: one with INLINE pragma and another with NOINLINE. It reads from vector so it should test changes to `basicUnsafeIndex`. I've run them with GHC 8.10, 9.2, 9.4, 9.6, 9.8
```haskell
varianceNoInline :: (VG.Vector v Double) => v Double -> Double
{-# NOINLINE varianceNoInline #-}
varianceNoInline xs
  = VG.sum (VG.map (\x -> (x - s)^(2::Int)) xs) / n
  where
    n = fromIntegral $ VG.length xs
    s = VG.sum xs / n
```

Function specialized by GHC runs in constant space. Changes to indexing make no difference and performance is stable for all GHC version with 6 cycles/element.

For non-specialized situation is much more interesting. Adding strictness (#485) make no difference al all in this benchmark. It seems programs are identical, al least number of instruction is exactly same, so I'll compare baseline (`stddev[base]`) and this PR (`stddev[unboxed]`). 

### Allocations
Below are allocations per array element:
```
  8.10.7     index[base]     96
   9.2.8     index[base]     96
   9.4.8     index[base]     96
   9.6.3     index[base]     96
   9.8.1     index[base]     96
  8.10.7  index[unboxed]     64
   9.2.8  index[unboxed]     64
   9.4.8  index[unboxed]     64
   9.6.3  index[unboxed]    128
   9.8.1  index[unboxed]    128
```
This PR is clear improvement for GHC<=9.4 but somehow it become a pessimization for GHC>=9.6. I haven't looked into core so I have no idea why

### Performance
Runtime performance follows same pattern:
```
    ghc             tag  INS/element  CYC/element
 8.10.7     index[base]  155.5        44.6
 8.10.7  index[unboxed]  128.6        32.7
  9.2.8     index[base]  153.9        44.9
  9.2.8  index[unboxed]  127.2        33.6
  9.4.8     index[base]  153.9        44.1
  9.4.8  index[unboxed]  127.3        34.1
  9.6.3     index[base]  189.9        65.0
  9.6.3  index[unboxed]  372.6       125.6
  9.8.1     index[base]  189.9        60.8
  9.8.1  index[unboxed]  372.6       120.3
```
25% win for GHC<=9.4 and 100% loss for GHC>=9.6. And latter performs worse even without this optimization. It looks we have some regression in GHC optimizer.

But I have looked into core yet and answer lies there.

PR #493, opened 2024-05-03
CI: bump to GHC 9.6.5; fix to macos-13 for GHC < 9.2

andreasabel:
macos-latest does not support GHC < 9.2


Shimuuar:
Thank you! Merged

Issue #443, opened 2022-07-18
Documentation?


asarkar:
README refers to hackage, hackage refers to itself, eventually, there's no usage instructions or tutorials for one to get started.

Shimuuar:
Vector does lack high level documentation. And unless someone steps up and write it it's going to stay this way. So I mark this issue as help-wanted

AlexMaryW:
Hi, I'd like to help out here :) just one question: what would you like the README to cover? I was thinking of some introduction, code examples/usage instructions, differentiation between vector and array. 

Please note that this [tutorial on Haskell Wiki](https://wiki.haskell.org/Numeric_Haskell:_A_Vector_Tutorial) is currently linked on Hackage, but maybe we could also consider linking [this one](https://mmhaskell.com/data-structures/vector)? I think it's less overwhelming and more concise. It might be a bit pointless to write another tutorial just for the sake of it if there are some good and easy-to-follow tutorials already out there.

PR #501, opened 2024-08-01
Improve README

AlexMaryW:
Issue: https://github.com/haskell/vector/issues/443

Hi, I had a go at improving the README. Do let me know what you think. Thanks! :)

### DIFF
--- a/vector/README.md
+++ b/vector/README.md
@@ -1,6 +1,89 @@
 The `vector` package [![Build Status](https://github.com/haskell/vector/workflows/CI/badge.svg)](https://github.com/haskell/vector/actions?query=branch%3Amaster)
 ====================
 
-An efficient implementation of `Int`-indexed arrays (both mutable and immutable), with a powerful loop optimisation framework.
+This package includes various modules that will allow you to 
+work with vectors and use an optimisation framework called [*fusion*](#fusion). 
+In this context, vector is an `Int`-indexed array-like data structure with a simpler 
+API that can contain any Haskell value. Additionally, its equivalence 
+to C-style arrays and optimisation via fusion accelerates vector’s 
+performance and makes it a great alternative to list. By installing this 
+package, you’ll be able to work with [boxed, unboxed, storable, and primitive 
+vectors](#vectors-available-in-the-package) as well as their generic interface.
 
-See [`vector` on Hackage](http://hackage.haskell.org/package/vector) for more information.
+
+## Table of Contents
+
+<!-- no toc -->
+- [Installation](#installation)
+- [Tutorial](#tutorial)
+- [Vector vs Array](#vector-vs-array)
+- [Vectors Available in the Package](#vectors-available-in-the-package)
+- [Fusion](#fusion)
+
+## Installation
+
+If you use **cabal**, modify `package.cabal` so that its `build-depends` 
+section includes vector package:
+```
+build-depends: base ^>=4.17.2.1
+             , vector ==0.13.1.0
+```
+If you use **stack**, modify `package.yaml` so that its `depends` 
+section includes vector package:
+```
+dependencies:
+- base >= 4.7 && < 5
+- vector == 0.13.1.0
+```
+
+
+## Tutorial
+
+A beginner-friendly tutorial for vectors can be found on 
+[MMHaskell](https://mmhaskell.com/data-structures/vector).
+
+
+If you have already started your adventure with vectors, 
+the tutorial on [Haskell Wiki](https://wiki.haskell.org/Numeric_Haskell:_A_Vector_Tutorial) 
+covers more ground.
+
+## Vector vs Array
+
+Arrays are data structures that can store a multitude of elements 
+and allow immediate access to every one of them. Even though Haskell 
+has a built-in [Data.Array module](https://hackage.haskell.org/package/array-0.5.7.0), 
+arrays might be a bit overwhelming to use due to their complex API. 
+Conversely, vectors incorporate the array’s *O(1)* access to elements 
+with a much friendlier API of lists. Since they allow for framework 
+optimisation via loop fusion, vectors emphasise efficiency and keep 
+a rich interface. Unless you’re confident with arrays, it’s 
+well-advised to use vectors when looking for a similar functionality.
+
+## Vectors Available in the Package
+
+**Boxed vectors** store each of its elements as a pointer to its value. 
+Because we cannot directly access the contents of a boxed vector, they 
+are slower in comparison to unboxed vectors.
+
+
+**Unboxed vectors** store solely their elements’ values instead of pointers. 
+To be unboxed, the elements need to be constant in size. Since we can directly 
+access the contents of the unboxed vector, working with them is quite efficient.
+
+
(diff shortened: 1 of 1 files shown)


vector/README.md
```diff
@@ -1,6 +1,89 @@
...
+the end of value and/or type’s name, e.g. `3#` or `Int#`. You can read 
+more about them [here](https://downloads.haskell.org/~ghc/5.00/docs/set/primitives.html).
+
+## Fusion
+
+An optimisation framework provided in this package, fusion 
+is a technique that merges several functions into one and forces 
+it to produce only one outcome. Without fusion, your program might 
+generate intermediate results for each function separately and 
+stall its performance.
```
Shimuuar, line 89: Specifically stream fusion as opposed to build/foldr fusion used for lists. 

```suggestion
Vector uses stream fusion for optimizations. This is technique 
allows to avoid creation of intermediate data structures. For example
following expression `sum . filter g . map f` will not allocate temporary vectors 
if compiled with optimizations.
```

vector/README.md
```diff
@@ -1,6 +1,89 @@
...
+with a much friendlier API of lists. Since they allow for framework 
+optimisation via loop fusion, vectors emphasise efficiency and keep 
+a rich interface. Unless you’re confident with arrays, it’s 
+well-advised to use vectors when looking for a similar functionality.
+
+## Vectors Available in the Package
+
+**Boxed vectors** store each of its elements as a pointer to its value. 
+Because we cannot directly access the contents of a boxed vector, they 
+are slower in comparison to unboxed vectors.
```
Shimuuar, line 66:
```suggestion
**Lazy boxed vectors** (`Data.Vector`) can store any haskell value. Their elements are
stored as pointers to heap-allocated values and that extra indirection results in worse
performance.

**Strict boxed vectors** (`Data.Vector.Strict`) boxed vector which is strict in its elements. 
```

vector/README.md
```diff
@@ -1,6 +1,89 @@
...
+
+
+**Storable vectors** are pinned, convertible to and from pointers, and 
+usable in C functions.
+
+
+**Primitive vectors** contain elements of primitive type. 
+Primitive types can be recognised by the hash sign attached at 
+the end of value and/or type’s name, e.g. `3#` or `Int#`. You can read 
+more about them [here](https://downloads.haskell.org/~ghc/5.00/docs/set/primitives.html).
```
Shimuuar, line 81: It's more precise to say that they can hold values that are instances of `Prim`. In fact it can't hold `Int#` — wrong kind.

```suggestion
**Primitive vectors** (`Data.Vector.Primitive`). This vector is backed by simple 
byte array and can hold data types which are instances of `Prim` type class.
This is data types which are represented in memory as sequence of bytes without
pointer. Think `Int`, `Double`, etc. Usually it's better to use unboxed vectors since
they have same performance and more general.
```

vector/README.md
```diff
@@ -1,6 +1,89 @@
...
+## Vectors Available in the Package
+
+**Boxed vectors** store each of its elements as a pointer to its value. 
+Because we cannot directly access the contents of a boxed vector, they 
+are slower in comparison to unboxed vectors.
+
+
+**Unboxed vectors** store solely their elements’ values instead of pointers. 
+To be unboxed, the elements need to be constant in size. Since we can directly 
+access the contents of the unboxed vector, working with them is quite efficient.
```
Shimuuar, line 71: In this case name lies.

```suggestion
**Unboxed vectors** (`Data.Vector.Unboxed`) is vector which determines representation
of an array from type of its element. For primitives (`Int`, `Double`, etc) it's backed by primitive
arrays and for tuples and product types by structure of arrays.  Generally it uses unboxed 
representation so it's quite efficient
```

vector/README.md
```diff
@@ -1,6 +1,89 @@
...
+
+
+**Storable vectors** are pinned, convertible to and from pointers, and 
+usable in C functions.
```
Shimuuar, line 75:
```suggestion
**Storable vectors** (`Data.Vector.Storable`) are backed by pinned memory. Their primary
use case is C FFI.  
```

Shimuuar (commented):
Thanks!

After many years of waiting someone improved rather anemic readme. I added suggestions which basically rewrote description of vectors.

And sorry for the wait.

AlexMaryW:
I read through and adapted your suggestions into the README + deleted the installation section.

vector/README.md
```diff
@@ -1,6 +1,72 @@
...
 
-An efficient implementation of `Int`-indexed arrays (both mutable and immutable), with a powerful loop optimisation framework.
+This package includes various modules that will allow you to 
+work with vectors and use an optimisation framework called [*fusion*](#fusion). 
+In this context, vector is an `Int`-indexed array-like data structure with a simpler 
+API that can contain any Haskell value. Additionally, its equivalence 
+to C-style arrays and optimisation via fusion accelerates vector’s 
+performance and makes it a great alternative to list. By installing this 
+package, you’ll be able to work with [boxed, unboxed, storable, and primitive 
+vectors](#vectors-available-in-the-package) as well as their generic interface.
```
Shimuuar, line 11:
```suggestion
An efficient implementation of `Int`-indexed arrays (both mutable and immutable).
This package provides multiple representations: [boxed, unboxed, storable, and primitive 
vectors](#vectors-available-in-the-package) and generic API which is polymorphic in
vector type. It also has powerful optimization framework based on [stream-fusion](#fusion)
which could eliminate intermediate data structures.
```
I think it would be better to cut down this paragraph a bit.

vector/README.md
```diff
@@ -1,6 +1,72 @@
...
+
+A beginner-friendly tutorial for vectors can be found on 
+[MMHaskell](https://mmhaskell.com/data-structures/vector).
+
+
+If you have already started your adventure with vectors, 
+the tutorial on [Haskell Wiki](https://wiki.haskell.org/Numeric_Haskell:_A_Vector_Tutorial) 
+covers more ground.
+
+## Vector vs Array
```
Shimuuar, line 32: Do we even need to compare with `array`. I think at this point it's mostly legacy and barely used

Shimuuar (commented):
I finally got review README again. Few more comments

@lehins @Bodigrim do you any input? After this PR is merged I want to start preparing long overdue 0.13.2.0 release

vector/README.md
AlexMaryW: Hi, I cut down the intro, but kept the sentence about vector's benefits (I think it's important to highlight them for newcomers)

vector/README.md
AlexMaryW: we can remove it, but I think we should keep it for the newcomers. If array is legacy and rarely used, I think it's important to mention that so I added a sentence

vector/README.md
Shimuuar: I dropped this: 

> The advantages of vectors include equivalence to C-style arrays and simple interface

because it's not clear what does "equivalence to C-style arrays" even mean?

AlexMaryW: removed it

Shimuuar:
I think it's  time to merge this PR. Thank you @AlexMaryW!

PR #508, opened 2024-10-10
A desirable Boxed instance for Unboxed

recursion-ninja:
This PR contains a proof of concept for `Unbox` instances of "boxed" data-types by way of the newly added `AsBoxedStrictly` and `AsBoxedLazily` newtypes. This PR is intended to address issue #503.

The `AsBoxedStrictly` newtype imposes new strictness semantics on the underlying type, ensuring the wrapped value is always reduced to normal form. This invariant requires an `NFData` instance for the underlying type. Furthermore, it requires the use of a "smart constructor" which evaluates the wrapped value to normal form, since a new-type cannot have strictness annotations. Hence there are `makeBoxedStrictly` and `grabBoxedStrictly` exposed from the `Data.Vector.Unboxed` module to construct and deconstruct `AsBoxedStrictly` values, respectively.

The `AsBoxedLazily` new-type is much simpler since it preserves the strictness semantics of the underlying data-type unaltered.

Here are two example usages of the newtypes as a proof of concept:

### **Strict usage**

```
data Foo a = Foo Int a
  deriving (Eq, Ord, Show)

instance NFData a => VU.IsoUnbox (Foo a) (Int, AsBoxedStrictly a) where
  toURepr (Foo i a) = (i, makeBoxedStrictly a)
  fromURepr (i, a) = Foo i $ grabBoxedStrictly a
  {-# INLINE toURepr #-}
  {-# INLINE fromURepr #-}
  
newtype instance VU.MVector s (Foo a) = MV_Foo (VU.MVector s (Int, AsBoxedStrictly a))

newtype instance VU.Vector    (Foo a) = V_Foo  (VU.Vector    (Int, AsBoxedStrictly a))

deriving via (Foo a `VU.As` (Int, AsBoxedStrictly a)) instance NFData a => VGM.MVector VUM.MVector (Foo a)

deriving via (Foo a `VU.As` (Int, AsBoxedStrictly a)) instance NFData a => VG.Vector   VU.Vector   (Foo a)

instance NFData a => VU.Unbox (Foo a)
```

In `GHCi`
```
import qualified Data.Vector.Unboxed as VU
VU.fromListN 3 [ Foo 4 "Hello", Foo 8 "there", Foo 16 "sailor" ]
[Foo 4 "Hello",Foo 8 "there",Foo 16 "sailor"]
```

### **Lazy usage**

```
data Bar a = Bar Int a
  deriving (Eq, Ord, Show)

instance VU.IsoUnbox (Bar a) (Int, AsBoxedLazily a) where
  toURepr (Bar i a) = (i, AsBoxedLazily a)
  fromURepr (i, AsBoxedLazily a) = Bar i a
  {-# INLINE toURepr #-}
  {-# INLINE fromURepr #-}
  
newtype instance VU.MVector s (Bar a) = MV_Bar (VU.MVector s (Int, AsBoxedLazily a))

newtype instance VU.Vector    (Bar a) = V_Bar  (VU.Vector    (Int, AsBoxedLazily a))

deriving via (Bar a `VU.As` (Int, AsBoxedLazily a)) instance VGM.MVector VUM.MVector (Bar a)

deriving via (Bar a `VU.As` (Int, AsBoxedLazily a)) instance VG.Vector   VU.Vector   (Bar a)

instance VU.Unbox (Bar a)
```

In `GHCi`
```
import qualified Data.Vector.Unboxed as VU
VU.fromListN 3 [ Bar 3 "Bye", Bar 2 "for", Bar 1 "now" ]
[Bar 3 "Bye",Bar 2 "for",Bar 1 "now"]
```

### *Nota Bene*

Since this is a proof of concept, all types and functions names are subject to change after the appropriate level of ["bike-shedding"](https://en.wiktionary.org/wiki/bikeshedding) has occurred. Commentary is welcomed and encouraged.

### DIFF
--- a/vector/src/Data/Vector/Unboxed/Base.hs
+++ b/vector/src/Data/Vector/Unboxed/Base.hs
@@ -27,11 +27,14 @@
 
 module Data.Vector.Unboxed.Base (
   MVector(..), IOVector, STVector, Vector(..), Unbox,
-  UnboxViaPrim(..), As(..), IsoUnbox(..)
+  UnboxViaPrim(..), As(..), IsoUnbox(..),
+  DoNotUnboxLazy(..), DoNotUnboxNormalForm(..), DoNotUnboxStrict(..)
 ) where
 
 import qualified Data.Vector.Generic         as G
 import qualified Data.Vector.Generic.Mutable as M
+import qualified Data.Vector                 as B
+import qualified Data.Vector.Strict          as S
 
 import qualified Data.Vector.Primitive as P
 
@@ -41,6 +44,7 @@ import Control.DeepSeq ( NFData(rnf)
 #if MIN_VERSION_deepseq(1,4,3)
                        , NFData1(liftRnf)
 #endif
+                       , force
                        )
 
 import Control.Monad.Primitive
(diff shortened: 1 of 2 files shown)


Shimuuar:
I haven't looked at the code yet. CI is broken at the moment fix should be in #507

vector/src/Data/Vector/Unboxed/Base.hs
```diff
@@ -764,6 +768,139 @@ instance (Unbox a, Unbox b) => G.Vector Vector (Arg a b) where
...
+-- >>> VU.fromListN 3 [ Foo 4 "Haskell's", Foo 8 "strong", Foo 16 "types" ]
+-- [ Foo 4 "Haskell's", Foo 8 "strong", Foo 16 "types" ]
+--
+-- @since 0.13.2.0n
+newtype AsBoxedStrictly a = AsBoxedStrictly a
+
+-- | Smart constructor for creating values of the 'AsBoxedStrictly' newtype.
+-- The wrapped values are evaluated to normal form via thier 'NFData' instance.
+--
+-- @since 0.13.2.0
```
Shimuuar, line 815: No reason to add constructor/accessor. Newtype's constructor should be exported anyway in order for deriving via to work.

 Anyway we aren't ensuring strictness in wrapper newtype. We're doing it in type class methods. Newtype is going to be coerced away. It's only used to pick instance.

Shimuuar (changes requested):
Please rebase on top current mater. CI is fixed there.

For strict and lazy vectors you simply need to pick `Data.Vector.Strict`/`Data.Vector` as a representation and coerce their methods same way as `UnboxViaPrim` do,

Variant which reduces values to NF is slightly more complicated. You can pick either strict or lazy vector as a representation. Former seems more logical. We evaluate value whenever it's stored in a vector. So following methods need custom code: `basicUnsafeWrite`,  `basicUnsafeReplicate`, `elemseq` 

Default implementation of `baiscSet` for boxed vectors `basicSet` is fine, it's defined in terms of `basicUnsafeWrite`

Shimuuar:
@recursion-ninja If you're trying to fix doctest it's easier to do so locally. Run in vector subdir.
```
$ cabal build all --write-ghc-environment-files=always
$ cabal run -- vector-doctest
```

recursion-ninja:
> @recursion-ninja If you're trying to fix doctest it's easier to do so locally. Run in vector subdir.
> 
> ```
> $ cabal build all --write-ghc-environment-files=always
> $ cabal run -- vector-doctest
> ```

Yes, I was. This is quite helpful.

vector/src/Data/Vector/Unboxed/Base.hs
recursion-ninja: I've updated the API accordingly.

recursion-ninja:
The test now pass, but I still need to do a pass over the documentation.

vector/src/Data/Vector/Unboxed/Base.hs
```diff
@@ -764,6 +767,264 @@ instance (Unbox a, Unbox b) => G.Vector Vector (Arg a b) where
...
+  basicLength          = coerce $ M.basicLength          @S.MVector @a
+  basicUnsafeSlice     = coerce $ M.basicUnsafeSlice     @S.MVector @a
+  basicOverlaps        = coerce $ M.basicOverlaps        @S.MVector @a
+  basicUnsafeNew       = coerce $ M.basicUnsafeNew       @S.MVector @a
+  basicInitialize      = coerce $ M.basicInitialize      @S.MVector @a
+  basicUnsafeReplicate = coerce $ M.basicUnsafeReplicate @S.MVector @a
+  basicUnsafeRead      = coerce $ M.basicUnsafeRead      @S.MVector @a
+  basicUnsafeWrite     = coerce $ M.basicUnsafeWrite     @S.MVector @a
+  basicClear           = coerce $ M.basicClear           @S.MVector @a
+  basicSet             = coerce $ M.basicSet             @S.MVector @a
```
Shimuuar, line 1006: We need to ensure that value is reduced to NF before writing it to vector. There are 3 method that could write to vector: `basicUnsafeReplicate`, `basicUnsafeWrite`, `basicSet`. They need custom definition which reduces element to NF before passing it to corresponding method

recursion-ninja: Thanks for this API insight. I have updated the PR `force`ing the values to normal form before being inserted into the vector.

recursion-ninja:
> The test now pass, but I still need to do a pass over the documentation.

I have updated the documentation.

@Shimuuar I think the PR is ready for review, and and potentially merged if the package maintainers agree.

vector/src/Data/Vector/Unboxed/Base.hs
```diff
@@ -764,6 +768,272 @@ instance (Unbox a, Unbox b) => G.Vector Vector (Arg a b) where
...
+  basicUnsafeNew       = coerce $ M.basicUnsafeNew       @S.MVector @a
+  basicInitialize      = coerce $ M.basicInitialize      @S.MVector @a
+  basicUnsafeReplicate i = coerce . (M.basicUnsafeReplicate @S.MVector @a) i . force . coerce
+  basicUnsafeRead      = coerce $ M.basicUnsafeRead      @S.MVector @a
+  basicUnsafeWrite v i x =
+    let y :: a
+        !y = force $ coerce x
+    in  M.basicUnsafeWrite v i $ coerce y
+  basicClear           = coerce $ M.basicClear           @S.MVector @a
+  basicSet v = coerce . M.basicSet v . force . coerce
```
Shimuuar, line 1011:
```suggestion
  basicUnsafeReplicate = coerce (\i x -> M.basicUnsafeReplicate @S.MVector @a i (force x))
  basicUnsafeRead      = coerce $ M.basicUnsafeRead      @S.MVector @a
  basicUnsafeWrite     = coerce (\v i x -> M.basicUnsafeWrite @S.MVector @a v i (force x))
  basicClear           = coerce $ M.basicClear           @S.MVector @a
  basicSet             = coerce (\v x -> M.basicSet @S.MVector @a v (force x))
```
Using `coerce` when more explicit pattern match would work it really poor style, it confuses reader. 

I think it would be better to stick to same style as rest of the instance: pick instance explicitly, then coerce it to corresponding newtypes

Shimuuar:
Other than comment above PR is good to go. 

I'm not sure why CI is failing. It seems to have stuck for some reason and then job got cancelled.

vector/src/Data/Vector/Unboxed/Base.hs
lehins: If I am not mistaken this implementation has an infinite loop, that's why CI is getting killed:
```
  basicUnsafeWrite v i x =
    let y :: a
        !y = force $ coerce x
    in  M.basicUnsafeWrite v i $ coerce y
```

The type of `v` doesn't change, so `basicUnsafeWrite` recursively calls itself.

Shimuuar: Indeed! And doctests do hang locally.

recursion-ninja: This is why we do the code reviews! Thanks for catching this.

recursion-ninja:
> Other than comment above PR is good to go.

I addressed the comment above. Should be good to merge now.

Shimuuar:
Merged. @recursion-ninja Thank you!

PR #512, opened 2024-10-30
Update documentation for unboxed vector

Shimuuar:
1. Drop NFData instance for DoNotUnboxNormalForm since it's not needed
2. Revert accidental deletion of documentation in 89d7584a9cb0b9aefe4676a497355ebc865c27de
3.  Rework overview of module for unboxed vectors
    
     - Mention that vector are not _necessarily_ unboxed
    
     - Enumerate all standard deriving methods. (And drop "Implementing unboxed
       vectors for new data types can be very easy", it always felt like mockery.

### DIFF
--- a/vector/src/Data/Vector/Unboxed.hs
+++ b/vector/src/Data/Vector/Unboxed.hs
@@ -13,14 +13,42 @@
 -- Stability   : experimental
 -- Portability : non-portable
 --
--- Adaptive unboxed vectors. The implementation is based on type families
+-- Adaptive unboxed vectors. The implementation is based on data families
 -- and picks an efficient, specialised representation for every element type.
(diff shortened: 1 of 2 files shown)


lehins (approved):
I don't really see a need to remove `NFData` instance. In fact it would make sense to derive it for all newtypes in vector.

That being said, considering that `DoNotUnboxNormalForm` is only used for deriving and NFData is not necessary for that then I have a problem with removing it.

Shimuuar:
Largely removal is for consistence. `DoNotUnboxNF` & friends are intended to be used only as instance selectors with deriving via. Other such wrappers don't have any more instances than required for for deriving. For example `UnboxViaPrim` doesn't have `Prim` instance.

PR #513, opened 2024-10-30
Drop "possibly" from force's haddocks

### LINKED ISSUE #511: Document that `force` always copies
The haddocks of the `force` function mentions that the input vector is *possibly* copied into a new return vector.

image

It seems however that `force` *always* copies regardless of the vector type, because the implementation for each vector type calls `Data.Vector.Generic.force`, which uses `basicUnsafeCopy`. 

If my understanding is correct, should the haddocks be updated to specify that the function always copies?

Shimuuar:
force indeed copies unconditionally. Even in case like:

> force (generate n fun)

In principle GHC could be able to elide copying using rewrite rules related to clone/New in practice it doesn't. So it's better to remove possibly since it will only confuse reader.

Fixes #511

### DIFF
--- a/vector/src/Data/Vector.hs
+++ b/vector/src/Data/Vector.hs
@@ -892,7 +892,7 @@ createT p = G.createT p
 -- ------------------------
 
 -- | /O(n)/ Yield the argument, but force it not to retain any extra memory,
--- possibly by copying it.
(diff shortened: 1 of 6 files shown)


lehins (approved):
:+1:

PR #518, opened 2024-12-19
Add Unboxed instances for Storable vectors

ruifengx:
This is a follow up for the `DoNotUnbox*` family of newtype `DerivingVia` wrappers (#508). It allows treating `Storable` vectors as unboxed vectors. My personal use case is combining parallel arrays received from C FFI into a single `Vector`. I left [a comment](https://github.com/haskell/vector/issues/503#issuecomment-2389978679) about this in #503, and did not get a response, but I still think this is a nice thing to have in `vector`.

I followed the implementation for `UnboxViaPrim` and named this new wrapper `UnboxViaStorable`. While working on this PR, I noticed that the documentation for `UnboxViaPrim` named the vector for `Foo` `[M]V_Int` instead of `[M]V_Foo`, so I also corrected them.

As a minor detail, the previous PR imports `Data.Vector.Strict` qualified as `S`, so I cannot use `S` for `Storable` and had to use `St`. I can change it if there is a preferred naming scheme for qualified imports.

### DIFF
--- a/vector/src/Data/Vector/Unboxed.hs
+++ b/vector/src/Data/Vector/Unboxed.hs
@@ -210,6 +210,7 @@ module Data.Vector.Unboxed (
   UnboxViaPrim(..),
   As(..),
   IsoUnbox(..),
+  UnboxViaStorable(..),
 
   -- *** /Lazy/ boxing
   DoNotUnboxLazy(..),
--- a/vector/src/Data/Vector/Unboxed.hs
+++ b/vector/src/Data/Vector/Unboxed.hs
@@ -68,7 +68,7 @@
 -- @
 module Data.Vector.Unboxed (
   -- * Unboxed vectors
-  Vector(V_UnboxAs, V_UnboxViaPrim), MVector(..), Unbox,
+  Vector(V_UnboxAs, V_UnboxViaPrim, V_UnboxViaStorable), MVector(..), Unbox,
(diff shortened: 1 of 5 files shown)


ruifengx:
I see some doctests errors and tried to fix them. Unfortunately, I cannot run doctests locally because it reports a lot of "Could not find module" errors, so I did it based on the logs from GitHub CI. After fixing the obvious typos, I am somewhat stuck right now because I do not quite understand how the module grouping in `doctests.hs` file works (I see that [only 8.10.7 and 9.0.2 are still failing](https://github.com/ruifengx/vector/actions/runs/12819635238/job/35747689988)).

Shimuuar:
I think PR is fine in general. We have support for deriving unboxed instances using primitive vectors, variants of boxed vector. No reason to not to add storable into the mix. 

I'll try to look why doctests fail on GHC 9.0,8.10

Shimuuar:
As it turns out failure was genuine. `V_UnboxViaStorable` must be in scope for deriving to work. Doctests just happen to work. I added tests to check that deriving actually works

ruifengx:
Thanks, I see the problem now. `doctest` "opens" the module before running (in the same way as `:m *SomeModule` in GHCi), making all internal definitions available, which presumably have made the tests pass for most compiler versions.

Issue #144, opened 2016-11-29
Can we get generateA?


treeowl:
```haskell
generateA :: Applicative f => Int -> (Int -> f a) -> f (Vector a)
-- specification
generateA n f = sequenceA (generate n f)
```

The `Monad m` constraint on `generateM` seems to flow from `unstreamM`. `unstreamM`, in turn, absolutely requires a `Monad m` constraint because of the way general streams are structured. For *generating* vectors, this all seems a bit sad, since the exact shape of the result stream is known in advance (a specific number of `Yield`s followed by `Done`). I don't know that there's any way around that annoyance.

Shimuuar:
Fixed by #522

Issue #132, opened 2016-07-28
Weaken constraint on mapM


andrewthad:
The type signatures of all of the monadic mapping functions (mapM, imapM, etc) have a `Monad` constraint on `m`. However, this can be weakened to `Applicative`. I do not know if this causes degradation of performance for certain `Monad`s though.

RyanGlScott:
I imagine using `Monad` here is intentional, since the `mapM` function from `Traversable` deliberately uses a `Monad` constraint. You're right that this could be weakened on `base-4.8` or later, but for backwards compatibility reasons it'd be prudent to keep around variants that use `Monad` rather than `Applicative`, since there is still some old code floating around that declares `Monad` instances but not `Applicative` ones. Similarly for `imapM`, `forM`, and `sequence`.

That being said, `Data.Traversable` _does_ contain `Applicative` counterparts to all of these functions (`sequenceA` vs. `sequence`, `traverse` vs. `mapM`, `for` vs. `forM`). It doesn't look like the `vector` library implements them directly (aside from [this inefficient implementation](https://github.com/haskell/vector/blob/16c263239c3b2d12454eba82b6b18041c8edaccc/Data/Vector.hs#L408)), so it might be prudent to add them.

Shimuuar:
It would require changes in stream fusion framework. Currently streams use Monad for sequencing of effects

andrewthad:
Thanks for the feedback from both of you. It seems like weakening the constraints on the existing functions would be bad for multiple reasons. Would introducing something like `itraverse`, which I am under the impression would be incompatible with the stream fusion framework, be alright? That is, is there a rule in place that everything that operates on immutable vectors must have a streaming rewrite, or do some function lack this?

cartazio:
It might not be possible because of the current fusion rep. If you can
weaken the internals to get by with applicative it may be possible

andrewthad:
@cartazio Sorry I wasn't clear. What I was asking is if it's alright to have a function that operates on immutable vectors that does not fuse.

cartazio:
Boxed vectors have traverse already right?
What would this applicative allow for the other formats?  I'm a tad slow
this morning so please expound.

andrewthad:
The original reason I brought this up was because the indexed traversal `imapM` has a `Monad` constraint. It would be nice to have `itraverse`, which would weaken that to `Applicative`.

RyanGlScott:
To be clear, I think there's two separate issues here:

[...]

Issue #69, opened 2015-01-23
add generic traverse method, a la Data.Traversable


cpeikert:
Unboxed vectors can't be Traversable due to the Unbox constraint on the element type, but it would be very useful if there were an analogous generic traverse method:

```
traverse :: (Applicative f, Vector v a, Vector v b) => (a -> f b) -> v a -> f (v b)
```

Even better would be if it were efficient, i.e., didn't just convert to and from lists (like the Traversable instance for boxed vectors does) -- but I'm not sure if this is even possible.

`sequence` comes close, except that it requires `Monad m` and `Vector v (m a)`, which will typically not hold for unboxed vectors.

cartazio:
GOOD IDEA.
I think we should totally get this into the 0.11 (we've a few other things we're keen on adding, and this totally makes sense to get into 0.11)

dolio:
It's actually difficult to be very efficient (I think). The problem is that the efficient way to build a vector is to write to a mutable vector and freeze. But when you are traversing, you have an arbitrary applicative involved blocking your ability to do efficient ST stuff. So, you can only build up ST actions as closures, or use an intermediate list/stream.

You can do better than creating a list first, of course. But it might always be a bit lackluster.

I think it's a good thing to have, though.

glguy:
I think it might look like this. Feel free to disregard the lens import, I was just using it for testing.

http://lpaste.net/119041

yongqli:
It would also be very useful to have `mapAccumL`, `mapAccumR`, `mapAccumLM`, `mapAccumRM`, etc.

cartazio:
Cant we define the stream / bundle version efficiently of traverse
efficiently though?

nh2:
Without contributing anything useful so far here, I'd benefit a lot from the `mapAccum*` functions for vectors.

ttuegel:
> Cant we define the stream / bundle version efficiently of traverse efficiently though?

*Maybe*. The trouble is that Stream and Bundle need an underlying monad, but traverse only gives us an applicative. But I *think* we can get away with a monad transformer that carries the applicative "along for the ride," as if it were some kind of monoidal functor.

Daniel-Diaz:
I am very interested in having some way of traversing vectors efficiently, especially in the way of the `mapAccum*` functions. Is this possible at all?

cartazio:
@dolio now that primMonad is stackable, can we do something for mutable vectors that partially addresses this?

cartazio:
[...]

PR #522, opened 2025-01-28
Implement Applicative based API 

### LINKED ISSUE #477: API using Applicative (traverse et.al.)
API based on `Applicative` has been requested multiple times:

 - #69 `traverse`
 - #132 weaken constraint on `mapM` (again `traverse`)
 - #144 `generateA`

This is quite natural request. `Applicative` is quite restrictive: we must know length of resulting vector in advance and we can't use result of one applicative value in other computations. This leaves us with four functions:

```haskell
replicateA :: Int -> f a          -> f (v a)
generateA  :: Int -> (Int -> f a) -> f (v a)
traverse   :: (a -> f b)        -> v a -> f (v b)
itraverse  :: (Int -> a -> f b) -> v a -> f (v b)
```

Did I forget some? But their implementation runs into several difficulties.

#### Problems

Applicatives are not compatible with stream fusion. Look at stream type:

```haskell
data Stream m a = forall s. Stream (s -> m (Step s a)) s

data Step s a where
  Yield :: a -> s -> Step s a
  Skip  :: s -> Step s a
  Done  :: Step s a
```

In order to generate next element of stream `m` must be monad. Thus we can not write `traverse` in terms of it.

Another problem is functions must work with _any_ applicative. This includes functions, list so `f (v a)` may require creating multiple vector, may involve complicated sharing. So we have to use pure data structures. This means we have
either to go through lists or build chain of closures. AFAIR `massiv` used latter approach which was more performant.

#### Existing implementations

Existing implementation in `lens` goes through list. Maybe its possible to improve upon it.

### Proposal

We can't relax constraint of `mapM` and friends. So we should leave them as they are.  Instead we should add set of non-fusible functions which use `Applicative`. In fact we only need to implement `generateA` all other functions could be trivially expressed in terms of it.

Selecting optimal implementation will require benchmarks. So I want to improve our benchmarks situation first

Shimuuar:
Implementation follows plan described in #477 with `generateA :: Applicative f => Int -> (Int -> f a) -> f (v a)` as primitive. It is not subject to stream fusion.  

Writing `generateA`  is not difficult. Problem is doing it efficiently. Current benchmarks are (suggestions for more are very welcome):

1. Generate vector of random numbers using state monad
2. Generate vector using IO action
3. Compute sum of vector using lens
4. Map vector using lens

First and naive implementation uses list as intermediate data structure. Sum benchmark performs well in this case. Using `STA` instead brings map benchmark on par with explicit loop and produces slight (5-10%) improvements in state and IO benchmark)
```haskell
newtype STA v a = STA {  _runSTA :: forall s. Mutable v s a -> ST s (v a) }
```

Currently sum and map perform on par with explicit loop.  State gives 7x slowdown and 8x allocations, IO benchmark 4x slowdown and 8x allocations. We obviously can add rewrite rules for `IO`/`ST` but maybe there're more general optimizations.

-----
Fixes #477, #69, #132, #144

### DIFF
--- a/vector/src/Data/Vector/Generic.hs
+++ b/vector/src/Data/Vector/Generic.hs
@@ -147,7 +147,8 @@ module Data.Vector.Generic (
   iscanr, iscanr',
 
   -- * Applicative API
-  replicateA, generateA, traverse, itraverse,
+  replicateA, generateA, traverse, itraverse, forA, iforA,
+  traverse_, itraverse_, forA_, iforA_,
 
   -- * Conversions
 
@@ -215,7 +216,7 @@ import Data.Typeable ( Typeable, gcast1 )
 
 import Data.Data ( Data, DataType, Constr, Fixity(Prefix),
                    mkDataType, mkConstr, constrIndex, mkNoRepType )
-import qualified Data.Traversable as T (Traversable(mapM,traverse))
+import qualified Data.Traversable as T (Traversable(mapM))
 
 -- Length information
 -- ------------------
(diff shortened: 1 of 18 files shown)


vector/src/Data/Vector/Generic.hs
```diff
@@ -2641,6 +2644,55 @@ clone v = v `seq` New.create (
...
+
+
+-- | Execute the applicative action the given number of times and store the
+-- results in a vector.
+replicateA :: (Vector v a, Applicative f) => Int -> f a -> f (v a)
+{-# INLINE replicateA #-}
+replicateA n f = generateA n (\_ -> f)
+
+
+-- | Construct a vector of the given length by applying the monadic
```
konsumlamm, line 2670:
```suggestion
-- | Construct a vector of the given length by applying the applicative
```

vector/src/Data/Vector/Generic.hs
```diff
@@ -2641,6 +2644,55 @@ clone v = v `seq` New.create (
...
+-- action to each index.
+generateA :: (Vector v a, Applicative f) => Int -> (Int -> f a) -> f (v a)
+{-# INLINE generateA #-}
+generateA 0 _ = pure empty
+generateA n f = runSTA n <$> go 0
+  where
+    go !i | i >= n    = pure $ STA unsafeFreeze
+          | otherwise =  (\a (STA m) -> STA $ \mv -> M.unsafeWrite mv i a >> m mv)
+                     <$> f i
+                     <*> go (i + 1)
```
konsumlamm, line 2680: It's probably better to use `liftA2` here.

konsumlamm (commented):
What about `for`/`for_`/`traverse_` etc?

Shimuuar:
> What about for/for_/traverse_ etc?

Good point! I forgot about them. But they're simpler since in this case one isn't in business  of constructing vector so it simply reduces to fold.

vector/src/Data/Vector/Generic.hs
Shimuuar: Probably. But I'm not sure that `STA` will survive maybe some `New`-like variant will perform better

Shimuuar:
I think that's about as far as reasonably possible for improving performance. 

For many `PrimMonads` we would like to simply create buffer and just write to it. But it's not safe in general case, applicatives with backtracking can break referential transparency. And it's not clear how to express such rewrite rule and whether it's even possible. Also `unstreamM` suffers from same problem

Shimuuar:
I think PR is ready.

PR #539, opened 2025-09-02
Use type application in order to simplify rewrite rules for unstreamM

Shimuuar:
Same rewrite rule only written more compactly

Draft since I'm not comfortable merging this PR (if it works on older GHC) without some test which checks that rewrite rule actually fires for reasonable code.


Shimuuar:
This syntax requires at least GHC8.8 so this have to be postponed until we drop support for older GHC

Shimuuar:
Now that support for GHC<8.8 is dropped and #548 is merged this PR could be merged with some assurance that we aren't breaking anything

Issue #240, opened 2019-04-05
Add uniquely named type aliases for each Vector type to `Data.Vector.Generic`


merijn:
This would let people more easily write code dealing with multiple `Vector` types without needing to do a qualified import from a different module for each type, just to get the type in scope. Some of my code has 7 lines of `Vector` imports just to bring different types into scope. Having unique alias for each in `Data.Vector.Generic` would solve that issue

chessai:
I agree that this would be useful. Anyone else's opinions on whether or not they would think this is useful?

Bikeshedding: What is your opinion on the names? What are others' opinions on the names?

Once we've sorted that out, I think a PR would be great.

cartazio:
I’ll hash this out with merijn sometime later this summer.  It’s not a
public mural ;)

mckeankylej:
This is a public mural. This is a well used open source library. You should be ashamed of your behavior.

cartazio:
to be clear, i'd rather avoid large bikesheddings,

cartazio:
and do the person who posted to this ticket and deleted their comment: i'm the maintainer :P

cartazio:
strawman:

BoxedIVec/BoxedMVec -- Data.Vector/Data.Vector.Mutable
UnboxedIVec / UnboxedMVec -- Data.Vector.Unboxed
CStructIVec / CStructMVec -- Data.Vector Storable

the last one is because the Vector use of storable is really about Pinned arrays (on or off heap) via foreign pointers to interact with Arrays of C compatible struct style data (we could call it )

an alternative for unboxed could be SOAIVec and SOAMVec, but that just seems terrible, even though Struct of Array IS a correct term of them. 

theres some realted issues of the current punning between the Vector type class and Vector types,which wasn't the case in vector major versions 0.1 and 0.2, but came into play later

ekmett:
I'd err on the side of being explicit and recycling existing naming.

What about just using e.g 

```
    StorableVector / MStorableVector
    BoxedVector / MBoxedVector
    PrimitiveVector / MPrimitiveVector
```

They are more verbose, yes. But that way there isn't any name mangling. 

The only thing that has to have a "made up" name is `BoxedVector` as `Boxed` doesn't show up anywhere else.

cartazio:
Sounds viable to me.

PR #540, opened 2025-09-03
Export constructors for vectors from unsafe module

### LINKED ISSUE #171: Feature Request: Make underlying representations accessible
In the (binary-serialise-cbor)[https://github.com/well-typed/binary-serialise-cbor] repo, there's been some discussion about the difficulty of being able to implement `Serialise` for `Vector`s which use more efficient representations of vectors than providing an array of structures in the encoded form. In well-typed/binary-serialise-cbor#62 I had proposed making the underlying representation of `Vector`s accessible via an associated type (this isn't precise when compared to the current implementation but I hope gets the idea across):

```haskell
class Vector v a where
    type VRepr v a :: *
    toVRepr :: v a -> VRepr v a
    fromVRepr :: VRepr v a -> Maybe (v a)
    ... -- The rest of the current Vector class

instance Vector 

instance (Vector v a, Vector v b) => Vector v (a,b) where
    type VRepr v (a,b) = (Vector v a, Vector v b)
    toVRepr (V_M2 _ va vb) = (va,vb) -- AKA unzip
    ...
```

Which allows for defining instances for the primitive types and instances for things like `(Serialise a, Serialise b) => Serialise (a,b)` and we get the instances for `Serialise (a,b)`, `Serialise (Complex a)` etc. essentially for free.

This is something I've run into a few times, where I wanted to take advantage of the underlying array of structures representation of Vectors without needing to know exactly which constructor was being used.

This is (obviously) not a very thorough proposal, I'm making it to spark discussion, and I'll be happy if the result is "no, because ... impossible ...", or "yes, we'll do this and change the major version".

### LINKED ISSUE #492: MVector data instances are exposed from Unboxed and Unboxed.Mutable
https://github.com/haskell/vector/blob/6b8bbc3a75b40451d8d225e30c576dfe89121c49/vector/src/Data/Vector/Unboxed.hs#L61-L63

https://github.com/haskell/vector/blob/6b8bbc3a75b40451d8d225e30c576dfe89121c49/vector/src/Data/Vector/Unboxed/Mutable.hs#L17-L19

I was confused as to why `Vector` data instances are not exposed but `MVector` instances are.

Then I found https://github.com/haskell/vector/issues/49#issuecomment-62709758 which explains why exposing them is unsafe, which is fair. But then there is no reason to expose the `MVector` instances. This wasn't answered in #49.

I would expect this to consistent between `Vector` and `MVector`. Or perhaps this issue is known but the `MVector` instances are not hidden to avoid breakage?

### LINKED ISSUE #506: Proposal to deprecate contructors for data unboxed vectors for tuples
They are very dangerous and allow to violate memory safety, could be replaced with `zip#`/`unzip#`. And attempts to use them safely largely amounts to rewriting of  `zip#`/`unzip#`. So I propose to deprecate them without plan for removal

### LINKED ISSUE #517: Why is MVector's constructor exposed?
The [constructor of `MVector`](https://hackage.haskell.org/package/vector-0.13.2.0/docs/Data-Vector-Mutable.html#t:MVector) is exposed:

```.hs
data MVector s a  = MVector !Int !Int !(MutableArray s a)
```

This seems pretty odd, because the two `Int`s are the offset into the underlying `MutableArray` that the `MVector` starts, and the length of the `MVector`. both of which are used for bounds checking before deferring to `readArray` of a `MutableArray` via [`basicUnsafeRead`](https://hackage.haskell.org/package/vector-0.13.2.0/docs/src/Data.Vector.Mutable.html#line-144).

So I'm confused. Firstly, can changing those fields not lead to segfaults? And secondly, doesn't changing those fields have no real purpose, since there are better ways of obtaining the ends one wants via the safe API (or even the unsafe one, with functions explicitly named as "`unsafe`" if one really wants the performance)?

Or have I missed something, and there is a good reason for exposing those fields?

### LINKED ISSUE #524: Storable Vector constructor is not exported
which fordbids doing some stuff.

The MVector constructor is exported, so I can workaround stuff for now, but it's still not great.

### LINKED ISSUE #535: Proposal for handling of vector constructors
Currently we export constructors for mutable vectors since they're needed for defining immutable ones and don't export constructors for immutable vector. There's two problems: 1) access to constructors is very unsafe and we expose them by default for mutable vectors  2) they're occasionally useful and we don't for immutable ones. 

1. For each vector data type we create `Unsafe` module and define both mutable and immutable vectors there. Export both with appropriate disclaimer. This module would be a good place for unsafe conversion functions as well.
2. Hide constructors for mutable vectors and provide deprecated pattern synonyms with same name for backward compatibility.

This way we can both provide access to constructor when one needs them and keep them hidden in standard scenarios.

Shimuuar:
This is proof of concept implementation of #535 for primitive vectors. 

Idea work out nicely. There's one small complication. It makes sense to define unsafe function working on representation of a vector but we get name clashes: both mutable and pure primitive vectors defined `unsafeCast` so it pprobab;y makes sense to keep such function where they're currently defined

Reviews are very welcome.

Fixes #535, Fixes #524, Fixes #517,  Fixes #506, Fixes #492, Fixes #171


Shimuuar:
I think PR is ready.

Shimuuar:
@lehins Do you have any comments? If no I'm going to merge this PR next weekend

lehins:
Sorry, I did want to review this PR.
I was travelling this week, so didn't have time to do this. Give me a few days and I will make sure to review it

Shimuuar:
No problem. I worked on PR for 4 month  on and off. It can wait.

On second thought dropping D.V.Unboxed.Base was too risque. It's better to deprecate that module and reexport content of new D.V.Unboxed.Unsafe

vector/src/Data/Vector/Mutable/Unsafe.hs
```diff
@@ -0,0 +1,187 @@
...
+fromMutableArray marr =
+  let size = sizeofMutableArray marr
+  in MVector 0 size `liftM` cloneMutableArray marr 0 size
+
+-- | /O(n)/ Make a copy of a mutable vector into a new mutable array.
+--
+-- @since 0.12.2.0
+toMutableArray :: PrimMonad m => MVector (PrimState m) a -> m (MutableArray (PrimState m) a)
+{-# INLINE toMutableArray #-}
+toMutableArray (MVector offset size marr) = cloneMutableArray marr offset size
```
lehins: I am not sure I understand why these two functions are moved to the unsafe module?
They both allocate a new array and make a full copy using safe bounds. I can't see anything unsafe about them.

vector/src/Data/Vector/Primitive/Mutable/Unsafe.hs
```diff
@@ -0,0 +1,135 @@
...
+
+type role MVector nominal nominal
+
+-- | Mutable vectors of primitive types.
+data MVector s a = MVector {-# UNPACK #-} !Int                  -- ^ offset
+                           {-# UNPACK #-} !Int                  -- ^ length
+                           {-# UNPACK #-} !(MutableByteArray s) -- ^ underlying mutable byte array
+
+type IOVector = MVector RealWorld
+type STVector s = MVector s
```
lehins: I think these two synonyms can go back into the safe module. They are totally safe and aren't needed in this module

vector/src/Data/Vector/Primitive/Mutable.hs
```diff
@@ -70,99 +72,22 @@ module Data.Vector.Primitive.Mutable (
...
-    where
-      sz = sizeOf (undefined :: a)
-
-  {-# INLINE basicSet #-}
-  basicSet (MVector i n arr) x = setByteArray arr i n x
+
+pattern MVector :: Int -> Int -> MutableByteArray s -> MVector s a
+pattern MVector i j arr = U.MVector i j arr
+{-# COMPLETE MVector #-}
+{-# DEPRECATED MVector "Use constructor exported from Data.Vector.Primitive.Mutable.Unsafe" #-}
```
lehins: Love the pattern synonym. 

What do you think about creating another one directional pattern synonym that would allow continue pattern matching on vectors in a safe manner? Can't think of any good names for it though.

With respect to deprecation, we can add some links to the message so they can be followed on haddock. This syntax will actually create links for the type and the module
```suggestion
{-# DEPRECATED MVector "Use `U.MVector` exported from \"Data.Vector.Primitive.Mutable.Unsafe\"" #-}
```

vector/src/Data/Vector/Strict/Unsafe.hs
```diff
@@ -0,0 +1,308 @@
...
+fromArray arr = liftRnf (`seq` ()) vec `seq` vec
+  where
+    vec = Vector $ V.fromArray arr
+
+-- | /O(n)/ Convert a vector to an array.
+--
+-- @since 0.13.2.0
+toArray :: Vector a -> Array a
+{-# INLINE toArray #-}
+toArray (Vector v) = V.toArray v
```
lehins: Same with these guys. I don't think there is anything unsafe about them at all. They all preserve the necessary invariants. They force the elements when needed and make copies when slicing cannot be preserved.

vector/src/Data/Vector/Strict/Unsafe.hs
```diff
@@ -0,0 +1,308 @@
...
+-- | /O(1)/ Extract the underlying `Array`, offset where vector starts and the
+-- total number of elements in the vector. Below property always holds:
+--
+-- > let (array, offset, len) = toArraySlice v
+-- > v === unsafeFromArraySlice len offset array
+--
+-- @since 0.13.2.0
+toArraySlice :: Vector a -> (Array a, Int, Int)
+{-# INLINE toArraySlice #-}
+toArraySlice (Vector v) = V.toArraySlice v
```
lehins: This one I can believe, for example, that it can be unsafe, since it is possible that some part of the result array will be uninitialized

vector/src/Data/Vector/Primitive.hs
```diff
@@ -24,7 +25,7 @@
 
 module Data.Vector.Primitive (
   -- * Primitive vectors
-  Vector(..), MVector(..),
+  Vector, MVector, pattern MVector, pattern Vector,
```
lehins: It's actually possible to export pattern synonyms together with types. This will minimize breakage for anyone who did `import Data.Vector.Primitive (Vector(..))` before
```suggestion
  Vector(MVector), MVector(Vector),
```

vector/src/Data/Vector/Primitive.hs
```diff
@@ -163,145 +164,24 @@ module Data.Vector.Primitive (
...
-  fromList = fromList
-  fromListN = fromListN
-  toList = toList
+  ( Eq, Ord, Num, Enum, Monoid, Traversable, Monad, Bool, Ordering(..), Int, Maybe, Either
+  , (==))
 
+pattern Vector :: Int -> Int -> ByteArray -> Vector a
+pattern Vector i j arr = U.Vector i j arr
+{-# COMPLETE Vector #-}
+{-# DEPRECATED Vector "Use constructor exported from Data.Vector.Primitive.Unsafe" #-}
```
lehins: Same here
```suggestion
{-# DEPRECATED Vector "Use `U.Vector` constructor exported from \"Data.Vector.Primitive.Unsafe\"" #-}
```

vector/src/Data/Vector/Strict/Mutable.hs
```diff
@@ -76,52 +78,21 @@ module Data.Vector.Strict.Mutable (
...
-  basicUnsafeCopy = coerce (G.basicUnsafeCopy @MV.MVector @a)
-
-  {-# INLINE basicUnsafeMove #-}
-  basicUnsafeMove = coerce (G.basicUnsafeMove @MV.MVector @a)
-  {-# INLINE basicClear #-}
-  basicClear = coerce (G.basicClear @MV.MVector @a)
+pattern MVector :: MV.MVector s a -> MVector s a
+pattern MVector v = U.MVector v
+{-# COMPLETE MVector #-}
+{-# DEPRECATED MVector "Use constructor exported from Data.Vector.Strict.Unsafe" #-}
```
lehins: Ditto
```suggestion
{-# DEPRECATED MVector "Use `U.MVector` constructor exported from \"Data.Vector.Strict.Unsafe\"" #-}
```

vector/src/Data/Vector/Unsafe.hs
```diff
@@ -0,0 +1,340 @@
...
+fromArray arr = Vector 0 (sizeofArray arr) arr
+
+-- | /O(n)/ Convert a vector to an array.
+--
+-- @since 0.12.2.0
+toArray :: Vector a -> Array a
+{-# INLINE toArray #-}
+toArray (Vector offset len arr)
+  | offset == 0 && len == sizeofArray arr = arr
+  | otherwise = cloneArray arr offset len
```
lehins: Same here. These conversions are totally safe

lehins (changes requested):
Aside from some functions unfairly being tainted as unsafe, I believe this is a very nice start for separating safe vs unsafe interface

Shimuuar:
@lehins Yes I moved some safe functions to `Unsafe` module. I found that when all function which work on vector representation directly and not via `Vector/MVector` type classes are defined in one place, it  improves readability. And there's little harm in exporting safe function from unsafe module. Similarly `Unsafe` module for unboxed vector has a lot of stuff that's quite safe. 

Real question is what should go into `Unsafe` modules and what should be eventually removed from safe ones? 

- Right now all function that work on vector representation directly are in defined in unsafe module. I think it's right choice
- Should we deprecate reexport of unsafe functions?
- Should we move unsafe indexing function (`unsafeIndex` etc) to unsafe modules and deprecate it as well?
- If so should we add `D.V.Generic.Unsafe` as well for unsafe indexing?


> What do you think about creating another one directional pattern synonym that would allow continue pattern matching on vectors in a safe manner? Can't think of any good names for it though.

What about `PrimVector`/`PrimMVector` etc?

lehins:
> Yes I moved some safe functions to Unsafe module. I found that when all function which work on vector representation directly and not via Vector/MVector type classes are defined in one place, it improves readability.  And there's little harm in exporting safe function from unsafe module.

I do not agree with this at all. This will make people wrongfully believe that those functions are unsafe. And I can certainly see harm in it!

> Real question is what should go into Unsafe modules and what should be eventually removed from safe ones?

Anything that is unsafe should IMHO be eventually moved: `unsafeIndex`, `unsafeFreeze`, etc. It should not be only about access to constructors, it should be about safe vs unsafe functionality.

> Right now all function that work on vector representation directly are in defined in unsafe module. I think it's right choice

Totally disagree. There is nothing wrong in providing a safe function that itself uses unsafe constructor.

> Should we deprecate reexport of unsafe functions?

Yes, I believe so.

> Should we move unsafe indexing function (unsafeIndex etc) to unsafe modules and deprecate it as well?

Yes, I believe so as well

> If so should we add D.V.Generic.Unsafe as well for unsafe indexing?

I do believe that would be a good idea for consistency and safety sake

> What about PrimVector/PrimMVector etc?

Doesn't sound too bad. Something like this?
* `PrimVector`/`PrimMVector`
* `StorableVector`/StorableMVector`
* `BoxedLazyVector`/`BoxedLazyMVector`
* `BoxedStrictVector`/`BoxedStrictMVector`

If these seem too long maybe we could shorten them:
* `PVector`/`PMVector`
* `SVector`/SMVector`
* `BLVector`/`BLMVector`
* `BSVector`/`BSMVector`

We could also provide matching type synonyms, so users could just import the type synonyms and use Generic interface if they so choose.

Shimuuar:
It turns that export of `Vector(Vector)` does not pick pattern defined in module, it packs imported constructor instead. But when module export synonym exporting `Mvector(MVector)` works fine. It seems only way to keep `import ... MVector(..)` working is to define deprecated patterns in separate module.

Also haddock doesn't link properly `'U.Vector'` and requires full qualification.

Shimuuar:
> I do not agree with this at all. This will make people wrongfully believe that those functions are unsafe. And I can certainly see harm in it!

So we disagree on this but your approach is valid as well and I don't feel to strong about his. Let do it your way. This PR is uncomfortably big already so I propose following plan:

1. Move safe functions back to safe module and merge this PR. It's good intermediate point
2. Next PR implements deprecation of unsafe conversion functions 
3. Third PR move unsafe indexing into `Unsafe` modules and deprecates their use from safe modules

> We could also provide matching type synonyms, so users could just import the type synonyms and use Generic interface if they so choose.

BTW exactly this was requested in #240 but I think it's better to do this separately

lehins:
I 100% agree with this plan: https://github.com/haskell/vector/pull/540#issuecomment-3877255972

Shimuuar:
OK step 1 is done. It turns out that supporting both qualified imports and import  `MVector(..)`/`MVector(MVector)` requires quite a bit of jumping through flaming hoops


    These patterns are used to deprecate import of wildly unsafe
    constructors from safe modules. There are two way in which they
    could be imported by user:
     > import Data.Vector...Mutable qualified
     > import Data.Vector...Mutable (MVector(..))
    and we want to support both.
    
    But there's a problem. When pattern is exported separately
     > , MU.MVector
     > , pattern MVector
    
    `import ... MVector(..)` doesn't work. It can't find constructor.
    Separate pattern doesn't count.
    
    And if constructor is in scope then following export
     > , MU.MVector(MVector)
    
    will always export constructor even if it only visible qualified
    and there's pattern with same name.
    
    So there's only one way: use `MVector(MVector)` export, never allow
    unsafe constructor in scope in safe modules if they need to export
    pattern instead.
    
    This requires defining pattern and all safe functions which make
    use of constructor should be defined in separate unexported module.

Shimuuar:
@lehins could you take a look at latest changes? In the end ensuring proper backward compatibility turned out to be quite ugly (but necessary).

vector/src/Data/Vector/Mutable/Pat.hs
```diff
@@ -0,0 +1,35 @@
+{-# LANGUAGE PatternSynonyms #-}
+module Data.Vector.Mutable.Pat
```
lehins: I'd just call it `Internal` instead, since there is a good chance we will want to put other stuff here besides a single pattern synonym. Also `Pat` is a pretty ugly shortening of "Pattern", at least I assume that's what it is. 
```suggestion
module Data.Vector.Mutable.Internal
```

Alternatively we could rename it to
```suggestion
module Data.Vector.Mutable.Pattern
```
and document it as temporary module that is needed during the period of constructor deprecation.

I just want to make sure we have a concrete plan for it:
* either we retain such modules on permanent basis as internal
* or document it as temporary, so those modules do get removed eventually

I don't have a strong preference one way or another.

vector/src/Data/Vector/Mutable/Pat.hs
```diff
@@ -0,0 +1,35 @@
...
+fromMutableArray marr =
+  let size = sizeofMutableArray marr
+  in MU.MVector 0 size <$> cloneMutableArray marr 0 size
+
+-- | /O(n)/ Make a copy of a mutable vector into a new mutable array.
+--
+-- @since 0.12.2.0
+toMutableArray :: PrimMonad m => MVector (PrimState m) a -> m (MutableArray (PrimState m) a)
+{-# INLINE toMutableArray #-}
+toMutableArray (MU.MVector offset size marr) = cloneMutableArray marr offset size
```
lehins: I can't seem to find a reason why these two functions are defined here?
As I've mentioned in my previous review, they are totally safe and belong in `Data.Vector.Mutable` instead

vector/src/Data/Vector/Mutable/Unsafe.hs
```diff
@@ -0,0 +1,158 @@
...
+
+type role MVector nominal representational
+
+-- | Mutable boxed vectors keyed on the monad they live in ('IO' or @'ST' s@).
+data MVector s a = MVector { _offset :: {-# UNPACK #-} !Int
+                           -- ^ Offset in underlying array
+                           , _size   :: {-# UNPACK #-} !Int
+                           -- ^ Size of slice
+                           , _array  :: {-# UNPACK #-} !(MutableArray s a)
+                           -- ^ Underlying array
```
lehins: Considering that with this PR we are starting to export field accessors as well, we should give them some decent names, eg:

```suggestion
data MVector s a = MVector { unsafeOffset :: {-# UNPACK #-} !Int
                           -- ^ Offset in underlying array
                           , unsafeSize   :: {-# UNPACK #-} !Int
                           -- ^ Size of slice
                           , unsafeMutableArray  :: {-# UNPACK #-} !(MutableArray s a)
                           -- ^ Underlying mutable array
                           }
```

vector/src/Data/Vector/Primitive/Mutable/Unsafe.hs
```diff
@@ -0,0 +1,129 @@
...
+----------------------------------------------------------------
+-- Mutable
+----------------------------------------------------------------
+
+type role MVector nominal nominal
+
+-- | Mutable vectors of primitive types.
+data MVector s a = MVector {-# UNPACK #-} !Int                  -- ^ offset
+                           {-# UNPACK #-} !Int                  -- ^ length
+                           {-# UNPACK #-} !(MutableByteArray s) -- ^ underlying mutable byte array
```
lehins: For consistency:
```suggestion
data MVector s a = MVector { unsafeOffset :: {-# UNPACK #-} !Int
                             -- ^ Offset into the `unsafeMutableByteArray` in number of elements, not bytes
                           , unsafeSize :: {-# UNPACK #-} !Int
                             -- ^ Size of the mutable vector in number of elements, not bytes
                           , unsafeMutableByteArray :: {-# UNPACK #-} !(MutableByteArray s)
                             -- ^ Underlying mutable byte array
                           }
```

vector/src/Data/Vector/Primitive/Pat.hs
```diff
@@ -0,0 +1,57 @@
...
+{-# COMPLETE Vector #-}
+{-# DEPRECATED Vector "Use Vector constructor exported from \"Data.Vector.Primitive.Unsafe\"" #-}
+
+pattern MVector :: Int -> Int -> MutableByteArray s -> MVector s a
+pattern MVector i j arr = MU.MVector i j arr
+{-# COMPLETE MVector #-}
+{-# DEPRECATED MVector "Use MVector exported from \"Data.Vector.Primitive.Mutable.Unsafe\"" #-}
+
+
+-- NOTE: [Constructor deprecation]
```
lehins: I think we could have solved this weird ghc behavior in a much cleaner way by using different more descriptive names like `UnsafeMVector` and `UnsafeVector`.
But, it's technically not required, since vector is usually imported qualified and "Unsafe" will likely be depicted in qualified name in one way or another.

vector/src/Data/Vector/Primitive/Unsafe.hs
```diff
@@ -0,0 +1,175 @@
...
+----------------------------------------------------------------
+-- Immutable
+----------------------------------------------------------------
+
+type role Vector nominal
+
+-- | Unboxed vectors of primitive types.
+data Vector a = Vector {-# UNPACK #-} !Int       -- ^ offset
+                       {-# UNPACK #-} !Int       -- ^ length
+                       {-# UNPACK #-} !ByteArray -- ^ underlying byte array
```
lehins:
```suggestion
data Vector a = Vector { unsafeOffset :: {-# UNPACK #-} !Int
                         -- ^ Offset into `unsafeByteArray` in number of elements, not bytes
                       , unsafeSize :: {-# UNPACK #-} !Int
                         -- ^ Number of elements in number of elements, not bytes
                       , unsafeByteArray :: {-# UNPACK #-} !ByteArray
                         -- ^ Underlying byte array
                       }
```

vector/src/Data/Vector/Storable/Mutable/Unsafe.hs
```diff
@@ -0,0 +1,288 @@
...
+import Unsafe.Coerce
+
+#include "vector.h"
+
+
+type role MVector nominal nominal
+
+-- | Mutable 'Storable'-based vectors.
+data MVector s a = MVector {-# UNPACK #-} !Int
+                           {-# UNPACK #-} !(ForeignPtr a)
```
lehins:
```suggestion
data MVector s a = MVector { unsafeSize :: {-# UNPACK #-} !Int
                           -- ^ Number of elements in a mutable vector
                           , unsafeForeignPtr :: {-# UNPACK #-} !(ForeignPtr a)
                           -- ^ Underlying buffer as a `ForeignPtr`, which is allowed to be mutated
                           }
```

vector/src/Data/Vector/Storable/Unsafe.hs
```diff
@@ -0,0 +1,219 @@
...
+-- This is marginally safer than 'unsafeCast', since this function imposes an
+-- extra 'Coercible' constraint. This function is still not safe, however,
+-- since it cannot guarantee that the two types have memory-compatible
+-- 'Storable' instances.
+unsafeCoerceVector :: Coercible a b => Vector a -> Vector b
+unsafeCoerceVector = unsafeCoerce
+
+-- | 'Storable'-based vectors.
+data Vector a = Vector {-# UNPACK #-} !Int
+                       {-# UNPACK #-} !(ForeignPtr a)
```
lehins:
```suggestion
data Vector a = Vector { unsafeSize :: {-# UNPACK #-} !Int
                         -- ^ Number of elements in a mutable vector
                       , unsafeForeignPtr :: {-# UNPACK #-} !(ForeignPtr a)
                         -- ^ Underlying buffer as a `ForeignPtr`, which is __NOT__ allow to be mutated
                       }

File: vector/src/Data/Vector/Mutable/Pat.hs [hunk 18]
```diff
@@ -0,0 +1,35 @@
...
+pattern MVector i j arr = MU.MVector i j arr
+{-# COMPLETE MVector #-}
+{-# DEPRECATED MVector "Use MVector exported from \"Data.Vector.Mutable.Unsafe\"" #-}
+
+-- See NOTE: [Constructor deprecation]
+
+-- | /O(n)/ Make a copy of a mutable array to a new mutable vector.
+--
+-- @since 0.12.2.0
+fromMutableArray :: PrimMonad m => MutableArray (PrimState m) a -> m (MVector (PrimState m) a)
```
lehins: These technically could be defined in the regular module, but we'd have to disable `-Wno-deprecations` warning for that module.
I don't have a preference about this either, just throwing it out there as an option.

vector/src/Data/Vector/Strict/Mutable/Unsafe.hs
```diff
@@ -0,0 +1,58 @@
...
+import qualified Data.Vector.Mutable as MV
+
+import Prelude ( Monad(..), return )
+
+#include "vector.h"
+
+type role MVector nominal representational
+
+-- | Mutable boxed vectors keyed on the monad they live in ('IO' or @'ST' s@).
+newtype MVector s a = MVector (MV.MVector s a)
```
lehins: Accessor could potentially be useful
```suggestion
newtype MVector s a = MVector { unsafeLazyMVector :: MV.MVector s a }
```

vector/src/Data/Vector/Strict/Unsafe.hs
```diff
@@ -0,0 +1,259 @@
...
+import Data.Semigroup ( Semigroup(..) )
+
+import qualified Control.Applicative as Applicative
+import qualified Data.Foldable as Foldable
+import qualified Data.Traversable as Traversable
+
+import qualified GHC.Exts as Exts (IsList(..))
+
+-- | Strict boxed vectors, supporting efficient slicing.
+newtype Vector a = Vector (V.Vector a)
```
lehins: Ditto
```suggestion
newtype Vector a = Vector { unsafeLazyVector :: V.Vector a }
```

vector/src/Data/Vector/Unsafe.hs
```diff
@@ -0,0 +1,311 @@
...
+import qualified Data.Foldable as Foldable
+import qualified Data.Traversable as Traversable
+
+import qualified GHC.Exts as Exts (IsList(..))
+
+
+-- | Lazy boxed vectors, supporting efficient slicing.
+data Vector a = Vector {-# UNPACK #-} !Int
+                       {-# UNPACK #-} !Int
+                       {-# UNPACK #-} !(Array a)
```
lehins:
```suggestion
data Vector a = Vector { unsafeOffset :: {-# UNPACK #-} !Int
                         -- ^ Offset into `unsafeArray`
                       , unsafeSize :: {-# UNPACK #-} !Int
                       -- ^ Number of elements
                       , unsafeArray :: {-# UNPACK #-} !(Array a)
                       -- ^ Underlying array
                       }
```

lehins (approved):
Few minor suggestions, but otherwise looks great!!!
I love that it provides graceful transition, instead of causing bunch of breakage for downstream users.

vector/src/Data/Vector/Primitive/Pat.hs
Shimuuar: That's actually great idea! Adding `unsafe` to names is generally good idea, and we do have safe constructors (newtype wrapper for unboxed vectors). No name clashes means that both safe and unsafe module could be imported into same namespace. For unsafe functions it will allow for cheap migration: one can just add import from unsafe module without touching rest of code.

Shimuuar:
Fields accessors also double as documentation. Adding them was right. I'll check that every public vector type except unboxed vectors have them.

@lehins thank you for reviews. They're most useful

P.S. There 500 responses in github's CI in addition to genuine error.  Where github is going? In any case I'll fix everything later.

Shimuuar:
I went with option of using `UnsafeVector`/`UnsafeMVector` constructor names. It makes it possible to move pattern into `Unsafe` module. 

But that causes another problem. `{-# DEPRECATED MVector #-}` deprecated both type and data constructor. GHC user guide suggests `{-# DEPRECATED data MVector #-}` but it work only in GHC>=9.10. 

I'll try to fix this problem later. Worst possible case we'll have to define patterns in separate module.

Shimuuar:
And now it's done

It turns out that it's not possible to deprecate only data constructor in older GHC so pattern has to be defined in separate module. `*.Unsafe` won't work. I called it `*.Pattern` because we have `Data.Vector.Storable.Internal` which is used by `Unsafe` modules.

PR #548, opened 2025-09-14
Optimization test which use allocation

Shimuuar:
This approach is complementary to inspection tests. We want to compile some functions on vectors to tight non allocating loops and it's difficult to formulate property  in terms of inspection tests. It's however easy to measure allocation and check that we don't allocate anything except vector's buffer and some constant overhead

These test trip up if rewrite rules for `unstreamM` and `generateA` are removed. So this is good way to check that rewrite rules and other optimizations work as intended and make much less scary to touch them.


Shimuuar:
And sure enough it found problems with optimizations in older GHC

PR #557, opened 2026-01-12
Simplify rewrite rules for enumFromTo

Shimuuar:
This is a simplification of rewrite rules: use type application for compactness and group them all together.

Also I found that one of them is wrong. Rule clearly intended for 32-bit builds was enabled for 64-bit instead.

https://github.com/haskell/vector/blob/8c0039f96be60b07e2c30810fab0b96bbc9623fb/vector/src/Data/Vector/Fusion/Bundle/Monadic.hs#L949-L980

Compare
https://github.com/haskell/vector/blob/8c0039f96be60b07e2c30810fab0b96bbc9623fb/vector/src/Data/Vector/Fusion/Bundle/Monadic.hs#L894-L899

That made me think. We notionally support 32-bit platforms, we have CPP for them, but we don't even test buildability on CI. Does anyone has ideas how to add such tests?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,7 +65,7 @@ jobs:
       name: Cache ~/.cabal/store
       with:
         path: ${{ steps.setup-haskell-cabal.outputs.cabal-store }}
-        key: ${{ runner.os }}-${{ matrix.ghc }}-CACHE_V4
+        key: ${{ runner.os }}-${{ matrix.ghc }}-CACHE_V5
     # ----------------
     - name: "Install PAPI"
       run: |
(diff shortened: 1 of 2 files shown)


konsumlamm:
> That made me think. We notionally support 32-bit platforms, we have CPP for them, but we don't even test buildability on CI. Does anyone has ideas how to add such tests?

GitHub actions provide an i386 container, see for example https://github.com/Bodigrim/bitvec/blob/95cdcf0c046b94af284ab4e7903da4d435fbf5c5/.github/workflows/ci.yaml#L49-L67. It's also possible to emulate various architectures (see below that), but that is significantly slower. You could also consider testing the JS and WASM backends (see for example https://github.com/Bodigrim/bitvec/blob/95cdcf0c046b94af284ab4e7903da4d435fbf5c5/.github/workflows/ci.yaml#L96-L159).

Shimuuar:
And surely enough I did break build on 32 bit platform. 

GHC9.0 & 9.2 fail on CI again with identical symptoms. I reset cache again. It helped but it can break again. I'm not sure what to do about it in that case

I need to recheck `enumFromTo_big_int` and friends. Whole situation looks suspicious,

PR #559, opened 2026-01-25
Add build on 32-bit platform on CI

Shimuuar:
Cleanup of RULES for enumFromTo (#557) turned out more complicated that I thought. So let split CI update into  separate PR.

Setup for running on 32-bit platform is copied from bitvec

Also CI cache is reset again. GHC 9.0 is failing again with same symptoms


Shimuuar:
This won't do. Allocation tests catch missing rewrite rule for Int64 on 32-bit platform. It will have to be merged as part of #557

PR #561, opened 2026-05-03
Deprecate reexports of unsafe functions defined in Unsafe modules

Shimuuar:
This is partial (only `Primitive` modules) implementation of stage 2 of plan described in https://github.com/haskell/vector/pull/540#issuecomment-3877255972 Once we're happy with implementation for primitive we can replicate it for other unsafe functions.

Deprecation of reexports was introduced in GHC9.8 so we have to define new functions to deprecate. Haddocks for deprecated reexports looks like this:
image


Shimuuar:
I think this PR is ready for review. @lehins do you have any objections?

lehins (approved):
LGTM 👍

